PR #2, opened 2024-05-31
iOS Voice Assistant

ErisMik:


recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/Info.plist
```diff
@@ -0,0 +1,51 @@
...
+	<key>CFBundlePackageType</key>
+	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleVersion</key>
+	<string>1</string>
+	<key>LSRequiresIPhoneOS</key>
+	<true/>
+	<key>NSMicrophoneUsageDescription</key>
+	<string>For voice activity detection</string>
```
laves, line 24: For voice assistant

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/AudioPlayerStream.swift
```diff
@@ -0,0 +1,78 @@
+import Foundation
```
laves, line 1: Since the other source files have our copyright header, you might as well include it here too

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/ViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+You can download directly to your device or airdrop from a Mac.
+"""
+    @Published var modelLoadStatusText = modelLoadStatusTextDefault
+    @Published var enableLoadModelButton = true
+    @Published var showFileImporter = false
+    @Published var selectedModelUrl: URL?
+
+    @Published var enginesLoaded = false
+
+    static let statusTextDefault = "Say `Picovoice` to start!"
```
laves, line 52: Since we show this every time before transcription/prompting, it makes more sense to just omit the 'to start'
```suggestion
    static let statusTextDefault = "Say `Picovoice`!"
```

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/ViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+    public func unloadEngines() {
+        do {
+            try VoiceProcessor.instance.stop()
+            VoiceProcessor.instance.clearFrameListeners()
+            VoiceProcessor.instance.clearErrorListeners()
+        } catch {
+            DispatchQueue.main.async { [self] in
+                errorMessage = "\(error.localizedDescription)"
+            }
+        }
```
laves, line 144: Does the voice processor record even when we're generating the completion and doing TTS? If that's the case, I think it would be better to start it on wakeword detect and stop it after cheetah finishes. This way we're running fewer threads and closing the mic stream when we're not using it

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/ViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+                let keyword = try self.porcupine!.process(pcm: frame)
+                if keyword != -1 {
+                    DispatchQueue.main.async { [self] in
+                        statusText = "Listening..."
+                        chatText.append(Message(speaker: "You", msg: ""))
+                        chatState = .STT
+                    }
+                }
+            }
+            if chatState == .STT {
```
laves, line 296: else if?

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/ViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+        do {
+            if audioStream?.isPlaying ?? false {
+                return
+            }
+            if chatState == .WAKEWORD {
+                let keyword = try self.porcupine!.process(pcm: frame)
+                if keyword != -1 {
+                    DispatchQueue.main.async { [self] in
+                        statusText = "Listening..."
+                        chatText.append(Message(speaker: "You", msg: ""))
```
laves, line 291:
```suggestion
                        chatText.append(Message(speaker: "You:", msg: ""))
```

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/ViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+                try dialog!.addHumanRequest(content: chatText[chatText.count - 1].msg)
+
+                DispatchQueue.main.async { [self] in
+                    chatText.append(Message(speaker: "picoLLM", msg: ""))
+                }
+
+                let result = try picollm!.generate(
+                    prompt: dialog!.prompt(),
+                    completionTokenLimit: 128,
+                    streamCallback: streamCallback)
```
laves, line 197: you can add stop phrases here

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/ViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+
+    private var completionQueue = DispatchQueue(label: "text-stream-queue")
+    private var completionArray: [String] = []
+
+    private func streamCallback(completion: String) {
+        DispatchQueue.main.async { [self] in
+            completionQueue.async {
+                self.completionArray.append(completion)
+            }
+            chatText[chatText.count - 1].append(text: completion)
```
laves, line 177: since we have the list of stop phrases, I filter them out so that we don't print or speak them

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/ViewModel.swift
```diff
@@ -0,0 +1,335 @@
...
+        errorMessage = ""
+
+        enableGenerateButton = false
+
+        DispatchQueue.global(qos: .userInitiated).async { [self] in
+            do {
+                try dialog!.addHumanRequest(content: chatText[chatText.count - 1].msg)
+
+                DispatchQueue.main.async { [self] in
+                    chatText.append(Message(speaker: "picoLLM", msg: ""))
```
laves, line 191:
```suggestion
                    chatText.append(Message(speaker: "picoLLM:", msg: ""))
```

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/PicoLLMVoiceAssistantDemoApp.swift
```diff
@@ -0,0 +1,19 @@
...
+//  file accompanying this source.
+//  Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+//  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+//  specific language governing permissions and limitations under the License.
+//
+
+import SwiftUI
+
+@main
+struct PicoLLMVoiceAssistantDemoApp: App {
```
laves, line 13: I think these cookbook demos aren't supposed to be as branded and picoLLM focussed (see the readme in the llm-voice-assistant dir 'LLM-Powered Voice Assistant'). I think we should just call this demo LLM Voice Assistant, like the demo directory is named.

recipes/llm-voice-assistant/ios/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+2. Open the `PicoLLMCompletionDemo.xcworkspace` in XCode
+
+3. Replace `let ACCESS_KEY = "${YOUR_ACCESS_KEY_HERE}"` in the file [VieModel.swift](./PicoLLMVoiceAssistantDemo/ViewModel.swift) with your AccessKey obtained from [Picovoice Console](https://console.picovoice.ai/).
+
+4. Build and run the project on your device.
+
+5. Press the `Load Model` button and load the model file from your device's storage.
+
+6. Say "picovoice", then speak to the voice assistant!
```
laves, line 44:
```suggestion
6. Say "Picovoice", then speak to the voice assistant!
```

res/.lint/spell-check/.cspell.json
```diff
@@ -11,6 +11,18 @@
     }
   ],
   "ignorePaths": [
-    "**/__pycache__/**"
+    "**/__pycache__/**",
+
+    // ios
```
laves, line 16: I think you have to add swift codestyle as an action

laves (changes requested):

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/ViewModel.swift
ErisMik: Actually, on iOS it seems that stopping and starting voice processor is a fairly expensive call. It seems to interrupt audio playback which makes a poor user experience.

recipes/llm-voice-assistant/ios/LLMVoiceAssistantDemo/ChatView.swift
```diff
@@ -0,0 +1,115 @@
...
+
+struct ChatView: View {
+    @ObservedObject var viewModel: ViewModel
+
+    var body: some View {
+        let isError = viewModel.errorMessage.count > 0
+
+        ZStack {
+            VStack(alignment: .center) {
+                Text("picoLLM Voice Assistant Demo")
```
ksyeo1010, line 20:
```suggestion
                Text("Voice Assistant Demo")
```

ksyeo1010:
I am getting some `failed to register bundle identifier` errors, does this happened to you?

ErisMik:
> I am getting some `failed to register bundle identifier` errors, does this happened to you?

Yeah this happens with all our demos, you have to change the bundle identifier to something like `ai.picovoice.LLMVoiceAssistantDemo.Kwangsoo` or something on your machine.

recipes/llm-voice-assistant/ios/PicoLLMVoiceAssistantDemo/ViewModel.swift
laves: That's odd... it shouldn't be that expensive and it really shouldn't affect the audio playback since stopping the input stream should happen entirely before the output stream starts. 

State should be:
`recording start -> wake word -> stt -> recording stop -> llm generate -> tts -> ...back to beginning`

It kind of sends a weird message to devs looking at this if we keep the mic stream open when it is not using audio. Especially if they use it as a basis for a project where they introduce even more app logic to the llm/tts portion. If there is a way to keep it open only when we need it, that would be preferable.

ErisMik:
@laves Can't seem to reply to your comment. Fixed it using the callbacks correctly, we now only record when needed. Also fixed a bug with not flushing the warmupPCM if it never reaches the 1 second mark.

laves (approved):

PR #8, opened 2024-08-07
Llm voice assistant python pvspeaker

albho:


ksyeo1010 (approved):

recipes/llm-voice-assistant/python/main.py
```diff
@@ -59,72 +61,72 @@ def tps(self) -> float:
 def orca_worker(access_key: str, connection, warmup_sec: float, stream_frame_sec: int = 0.03) -> None:
     # noinspection PyUnresolvedReferences
     import numpy as np
```
kenarsa, line 63: can you remove this? I don't think it's needed?

recipes/llm-voice-assistant/python/main.py
```diff
@@ -59,72 +61,72 @@ def tps(self) -> float:
...
-    from sounddevice import OutputStream
```
kenarsa, line 62: Can we import it in the global scope? For `sounddevice` I couldn't because otherwise the device was owned by the wrong process.

recipes/llm-voice-assistant/python/main.py
```diff
@@ -9,6 +9,8 @@
     Optional,
     Sequence,
 )
+from collections import deque
```
kenarsa, line 12: can you use pycharm to make sure the order is correct?

PR #25, opened 2024-11-20
Picollm v1.2

matt200-ok:


recipes/llm-voice-assistant/python/main.py
```diff
@@ -299,17 +326,17 @@ def llm_task(dialog, user_request, utterance_end_sec, main_connection):
...
 
-        completion = ['']
+        completion = CompletionText(stop_phrases)
 
         def llm_callback(text: str) -> None:
             picollm_profiler.tock()
-            completion[0] += text
-            if not any(x in completion[0] for x in stop_phrases):
+            diff = completion.append(text)
+            if len(diff) > 0:
```
laves, line 334: to be honest, the CompletionText class feels a bit convoluted to me and the API is a bit weird (append returning a diff). Can we simplify this into just a couple more lines of code from the original?

PR #28, opened 2024-12-06
NodeJS interrupt fix

matt200-ok:


recipes/llm-voice-assistant/nodejs/package.json
```diff
@@ -16,7 +16,7 @@
   "dependencies": {
     "@picovoice/cheetah-node": "^2.0.2",
```
laves, line 17: we should update cheetah while we're at it

recipes/llm-voice-assistant/web/package.json
```diff
@@ -21,7 +21,7 @@
   "dependencies": {
     "@picovoice/cheetah-web": "^2.0.0",
```
laves, line 22: update cheetah

laves (changes requested):
the web demo when generating says `Generating <spinner> say Picovoice to interrupt` when generating. Let's do `<spinner> Generating... (say Picovoice to interrupt)` instead

laves:
also let's update web cheetah to latest as well

recipes/llm-voice-assistant/web/package.json
```diff
@@ -19,6 +19,7 @@
   "author": "Picovoice Inc",
   "license": "Apache-2.0",
   "dependencies": {
+    "@picovoice/cheetah-node": "^2.1.0",
```
laves, line 22: looks like you added cheetah node here instead of updating cheetah web?

laves (changes requested):

recipes/llm-voice-assistant/web/package.json
laves: remove and update the yarn lock so that it is not there

laves (approved):

PR #31, opened 2024-12-20
Python gui demo

matt200-ok:


laves (changes requested):
two additional things:
- non-windows should fail
- change source file name to `windows_gui.py
- when can't find `--config_path`, it should fail

laves (approved):

laves (approved):

PR #35, opened 2024-12-27
iOS interrupt fix

matt200-ok:


recipes/llm-voice-assistant/ios/Podfile
```diff
@@ -4,7 +4,7 @@ platform :ios, '16.0'
 target 'LLMVoiceAssistantDemo' do
   pod 'Porcupine-iOS', '~> 3.0.1'
   pod 'Cheetah-iOS', '~> 2.0.0'
```
laves, line 6: update cheetah

laves (changes requested):
move to SPM as we talked about

Also, the generating text is `Generating, say Picovoice to interrupt` all on one line - let's change that to match Android `Generating...\nsay Picovoice to interrupt` if that looks okay

recipes/llm-voice-assistant/ios/README.md
```diff
@@ -27,11 +27,7 @@ Copy your model file to the device.
...
 ## Usage
 
-1. Install the dependencies using `CocoaPods`:
-
-```console
-pod install
-```
+1. Install the dependencies using `Swift Package Manager`:
 
 2. Open the `LLMVoiceAssistantDemo.xcworkspace` in XCode
```
laves, line 32: Since SPM package install is automatic and there is no longer an xcworkspace file, I think we need to change this to `1. Open the LLMVoiceAssistantDemo.xcodeproj in XCode`

laves (changes requested):

laves (approved):

PR #37, opened 2025-01-16
Windows Arm64

matt200-ok:


laves (changes requested):
- cap UI at a max height and width
- cap resource usage bars at the same size as VU meter bars

laves (approved):

PR #38, opened 2025-02-21
.NET demo

picoliu:


matt200-ok (commented):
I think we should add a dotnet codestyle action for this demo.

matt200-ok (approved):

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,653 @@
...
+                Channel.CreateUnbounded<string>();
+            public event EventHandler<string>? PartialCompletionGenerated;
+            public event EventHandler<PicoLLMEndpoint>? CompletionGenerationCompleted;
+
+            public Generator(Listener listener)
+            {
+                _pllm = PicoLLM.Create(accessKey: ACCESS_KEY, modelPath: LLM_MODEL_PATH,
+                                       device: LLM_DEVICE);
+                listener.UserInputReceived += OnUserInputReceived;
+                listener.WakeWordDetected += OnWakeWordDetected;
```
laves, line 223: it looks like the only reason `listener` is being passed in to Generator and Speaker is to have some callbacks subscribed to it. That's not a great reason to hold on to an instance. Set these callbacks after you've created the Listener and Generator instead:
```csharp
Listener listener = new Listener();
Generator generator = new Generator();
listener.UserInputReceived += generator.OnUserInputReceived;
listener.WakeWordDetected += generator.OnWakeWordDetected;
```

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Profiler.cs
```diff
@@ -0,0 +1,6 @@
+using System.Data;
+
+interface Profiler
```
laves, line 3: Though it's not usually what we'd suggest, I think we're trying to keep all the code in one file for these demos so that people trying to learn don't get too confused. Can you move all the Profiler stuff and its subclasses into the main program file, like we have in Python and Node? You can put it at the bottom

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/LLMVoiceAssistant.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="picollm" Version="1.2.1" />
+    <PackageReference Include="Picovoice.Cheetah" Version="2.1.2" />
+    <PackageReference Include="Picovoice.Orca" Version="1.0.0" />
```
laves, line 13: Not sure if there's new versions of the others, but Orca is now `1.1.0`

laves (changes requested):
Some of the feedback is general that applies to many areas. Reread your code with them in mind and fix along the way.

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,653 @@
...
+                    try
+                    {
+                        orcaStream = _orca.StreamOpen(speechRate: ORCA_SPEECH_RATE);
+                        break;
+                    }
+                    catch (System.Exception)
+                    {
+                        continue;
+                    }
+                }
```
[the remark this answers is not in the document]
picoliu: We initialize a new orca stream every time the wake up word is said. If it's the normal situation, this should success in one go. But if it's triggered by an interruption, it has to wait for the previous run to finish and release the orca stream.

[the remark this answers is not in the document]
laves: I think the logic needs to be more specific to the situation then. If you're waiting on something to finish before init, you should have some sort of signal that you wait for and read before you attempt

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,653 @@
...
+                        }
+                    }
+                }).ContinueWith((t) =>
+                {
+                    speaker.Flush();
+                    Console.WriteLine($"Say {PORCUPINE_WAKE_WORD} to start.");
+                });
+                Task.Run(async () =>
+                {
+                    while (true)
```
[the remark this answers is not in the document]
picoliu: Most of the time this code will stuck at `await completionChannel.Reader.ReadAsync(cancellationToken);`, and when the cancellation is triggered, this call will throw an exception and break the loop automatically.

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,653 @@
...
+                        completionTokenLimit: COMPLETION_TOKEN_LIMIT,
+                        stopPhrases: STOP_PHRASES, presencePenalty: LLM_PRESENCE_PENALTY,
+                        frequencyPenalty: LLM_FREQUENCY_PENALTY,
+                        temperature: LLM_TEMPERATURE, topP: LLM_TOP_P,
+                        streamCallback: callback);
+                    dialog.AddLLMResponse(result.Completion);
+                    Console.WriteLine();
+                    CompletionGenerationCompleted?.Invoke(this, result.Endpoint);
+                    llmFinish.SetResult(true);
+                }
```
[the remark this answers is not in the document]
laves: I see, this is the body of the worker's task. Can you make Listener, Generator and Speaker have a more consistent interface? One uses `run`, one uses a `Task.Run` on init and the other uses an inline task call. I found that a bit hard to read

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,653 @@
...
+                cheetahProfiler = new RTFProfiler("Cheetah");
+                pllmProfiler = new TPSProfiler("PicoLLM");
+                delayProfiler = new DelayProfiler();
+            }
+            Listener listener = new Listener();
+            Generator generator = new Generator(listener);
+            Speaker speaker = new Speaker(generator, listener);
+            llmFinish.SetResult(true);
+            Console.WriteLine($"Initialized. Say {PORCUPINE_WAKE_WORD} to start.");
+            listener.run();
```
laves, line 607: Function names should be capitalized: `Run`

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,653 @@
...
+                            LLM_DEVICE = (llmDeviceObj as JsonElement?)?.GetString() ?? LLM_DEVICE;
+                            config.TryGetValue("picollm_completion_token_limit", out object? completionTokenLimitObj);
+                            COMPLETION_TOKEN_LIMIT = (completionTokenLimitObj as JsonElement?)?.GetInt32() ?? COMPLETION_TOKEN_LIMIT;
+                            config.TryGetValue("picollm_presence_penalty", out object? llmPresencePenaltyObj);
+                            LLM_PRESENCE_PENALTY = (llmPresencePenaltyObj as JsonElement?)?.GetSingle() ?? LLM_PRESENCE_PENALTY;
+                            config.TryGetValue("picollm_frequency_penalty", out object? llmFrequencyPenaltyObj);
+                            LLM_FREQUENCY_PENALTY = (llmFrequencyPenaltyObj as JsonElement?)?.GetSingle() ?? LLM_FREQUENCY_PENALTY;
+                            config.TryGetValue("picollm_temperature", out object? llmTemperatureObj);
+                            LLM_TEMPERATURE = (llmTemperatureObj as JsonElement?)?.GetSingle() ?? LLM_TEMPERATURE;
+                            config.TryGetValue("picollm_top_p", out object? llmTopPObj);
```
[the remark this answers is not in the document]
picoliu: The `??` operator does the null check, if the lhs variable is null, it will return the rhs variable. But creating a new class is a better idea.

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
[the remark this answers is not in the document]
laves: Sounds like you might as well check `!cancellationToken.IsCancellationRequested` then for the situations that don't fall into that

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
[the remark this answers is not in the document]
laves: the points of failure I was talking about were just getting null when we should have gotten a real value given how many nullable escapes there are here. Silent bugs are no fun

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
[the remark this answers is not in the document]
picoliu: If the cancellation is triggered, for example, while synthesizing, the next time we wait on `ReadAsync` will still trigger an exception.

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
picoliu: I just found out that events cannot be passed as parameters, and we do need to register and remove some handlers repeatedly in speaker, so at least in the speaker, it must hold a reference to the generator.
As for other cases, I feel like the subscriptions are a part of initialization - without them the generator and speaker cannot work correctly, therefore we should pass the listener (or the events, but unfortunately we can't) at initialization time. Besides, the listener and generator classes don't expose a lot of unnecessary functions, basically all public members are events, so holding a reference to generator/listener might not be a very bad idea.

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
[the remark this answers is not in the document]
picoliu: The listener's `Run` is to run on the main thread and block the `Main` from returning, since we have to run something on the main thread, I just chose that function. But I can refactor to make the main thread sleep and put that to another thread. In that case, `Listener` will have a similar behaviour as `Generator`.
As for the `Speaker`, it will create a orca synthesis task and a speaker output task every time the wake word is said. This design is to minimize synchronization complications by starting fresh for every LLM completion.

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,653 @@
...
+                    CompletionGenerationCompleted?.Invoke(this, result.Endpoint);
+                    llmFinish.SetResult(true);
+                }
+            }
+        }
+
+        private class Speaker
+        {
+            readonly Orca _orca;
+            readonly Generator generator;
```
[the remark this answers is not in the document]
picoliu: `readonly` is the constant modifier in C#, `const`s are compile-time constants.

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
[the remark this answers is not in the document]
laves: I'm commenting more on the consistency of the interfaces, not the behaviour of Listener, Generator and Speaker. At their core, they are all workers with main job loops. However, their main job loops are hard to find because they are all named different things and are declared using different conventions. Do what you can to bring them to a more similar looking interface to improve readability.

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
laves: There's pros and cons for forging class connections at the class-level. I think it's usually best to reduce dependencies between classes by letting application code manage connections. By linking Listener handlers and Generator events internally, you limit both classes. Having events be public is not a bad thing - it basically tells the developer that this class will fire events and if you want to subscribe to them, you can write a handler to do so.

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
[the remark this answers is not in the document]
laves: Please be sure items have been addressed before resolving a conversation. I was commenting on adding the explicit visibility to your `readonly` members, not changing from `readonly` to `private`. Adding `private` and being explicit with visibility is best practice. Also on second thought, the preceding `_` on private members is way more common to C# coders, so please add that to all of your private members.

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,751 @@
...
+            event EventHandler<short[]>? AudioReceived;
+            string transcript = "";
+            public event EventHandler<string>? UserInputReceived;
+            public event EventHandler? WakeWordDetected;
+
+            public Listener()
+            {
+                cheetah = Cheetah.Create(accessKey: config.AccessKey,
+                                   endpointDurationSec: config.CheetahEndpointDurationSec,
+                                   enableAutomaticPunctuation: true);
```
laves, line 130: multi-arg calls should have line break after the parenthesis:
```csharp
cheetah = Cheetah.Create(
        accessKey: config.AccessKey,
        endpointDurationSec: config.CheetahEndpointDurationSec,
        enableAutomaticPunctuation: true);
```

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,751 @@
...
+            }
+
+            private void CheetahOnAudioReceived(object? sender, short[] pcm)
+            {
+                cheetahProfiler?.Tick();
+                CheetahTranscript transcript = cheetah.Process(pcm);
+                cheetahProfiler?.Tock(pcm);
+                if (transcript.Transcript.Length > 0)
+                {
+                    this.transcript += transcript.Transcript;
```
laves, line 180: having to have `transcript` and `this.transcript` in the same block is a good indication that you should change one of them to reduce confusion. I recommend changing the local var `transcript` to `result` or something

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,751 @@
...
+                        break;
+                    default:
+                        argIndex++;
+                        break;
+                }
+            }
+
+            if (config.AccessKey == null || config.PicollmModelPath == null)
+            {
+                Console.WriteLine("Access key and LLM model path are required.");
```
laves, line 552: better to use the arg names
```suggestion
                Console.WriteLine("`access_key` and `picollm_model_path` are required.");

```

recipes/llm-voice-assistant/dotnet/README.md
```diff
@@ -0,0 +1,77 @@
...
+dotnet run -c Release --access_key ${ACCESS_KEY} --picollm_model_path ${PICOLLM_MODEL_PATH} 
+```
+
+Replace `${ACCESS_KEY}` with yours obtained from Picovoice Console and `${PICOLLM_MODEL_PATH}` with the path to the 
+model downloaded from Picovoice Console.
+
+To see all available options, type the following:
+
+```console
+./bin/Debug/net8.0/LLMVoiceAssistant --help
```
laves, line 49:
```suggestion
dotnet run -c Release --help
```

laves (changes requested):

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,793 @@
...
+
+namespace LLMVoiceAssistant
+{
+    class LLMVoiceAssistant
+    {
+        static RTFProfiler? orcaProfiler;
+        static RTFProfiler? porcupineProfiler;
+        static RTFProfiler? cheetahProfiler;
+        static TPSProfiler? pllmProfiler;
+        static DelayProfiler? delayProfiler;
```
laves, line 16: do all of these need to be static vars for the program? It looks like each worker class instance could instantiate their relevant profilers on init - you may just have to pass in the profile bool to the constructors

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
```diff
@@ -0,0 +1,793 @@
...
+
+    public RTFProfiler(string name)
+    {
+        Name = name;
+    }
+
+    public void Init(int sampleRate)
+    {
+        _sampleRate = sampleRate;
+    }
```
laves, line 683: If you create the profiler in the actual worker classes then you can change the constructor to 
```csharp
public RTFProfiler(string name, int sampleRate)
```
and remove the `Init` function

laves (changes requested):

recipes/llm-voice-assistant/dotnet/LLMVoiceAssistant/Program.cs
picoliu: Some profilers have to be used in different classes, for example, `delayProfiler`, making these profilers have to be defined in a larger scope. So I just create all profilers together.

laves: ones that need to be shared between classes can be global if need be, but it looks like 4/5 could move to be instance variables which would simplify profiler inits and localize them to where they're actually used

laves (approved):

PR #79, opened 2026-04-09
speech translation recipes

kenarsa:


recipes/live-conversation-translation/python/setup.py
```diff
@@ -0,0 +1,39 @@
+import os
+import shutil
+import subprocess
+
+
+def clone_repo(animal: str, major: str, minor: str) -> None:
```
laves, line 30: while this definitely offers some convenience, I think it doesn't offer an overall positive UX. The script runs for a while without progress indication and clones three whole repos, when the user just needs the model files. Ideally the user only downloads the files that are needed for the demo and they are kept up-to-date on the progress. Maybe we use the GitHub API to download the folder contents directly or use sparse checkout to target only what we need?

matt200-ok (approved):

recipes/live-conversation-translation/python/README.md
```diff
@@ -0,0 +1,65 @@
...
+[Orca Streaming Text-to-Speech](https://picovoice.ai/docs/orca/):
+
+```console
+python setup.py
+```
+
+### 5. Run the Demo
+
+```console
+python main.py --access_key ${ACCESS_KEY} --language_pair ${LANGUAGE_PAIR}
```
laves, line 61: I think we should give the user an example of what a language pair would look like. Maybe a line below this that says. 

```
Replace ${LANGUAGE_PAIR} with the pair of languages you would like to translate between, e.g. `en-fr` for English <> French translation. 
```

recipes/live-conversation-translation/python/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```console
+source .venv/bin/activate
+```
+
+On Windows:
+
+```console
+.venv\Scripts\activate
+```
```
laves, line 40: not against suggesting a virtual environment to users, but it doesn't make the steps to run this demo look quite a bit longer that they need to be. In the other demos in the repo I believe we just do:

Install the required packages:

```console
pip install -r requirements.txt
```

Run the demo:

```console
python3 main.py --access_key ${ACCESS_KEY} --picollm_model_path ${PICOLLM_MODEL_PATH} 
```

Replace `${ACCESS_KEY}` with yours obtained from Picovoice Console and `${PICOLLM_MODEL_PATH}` with the path to the 
model downloaded from Picovoice Console.

To see all available options, type the following:

```console
python main.py --help
```

ksyeo1010 (approved):

recipes/live-conversation-translation/python/main.py
```diff
@@ -0,0 +1,260 @@
...
+    parser.add_argument(
+        '--disable_text_normalization',
+        action='store_true',
+        help='Disable text normalization in Streaming Speech-to-Text.')
+    parser.add_argument(
+        '--genders',
+        nargs=2,
+        choices=['female', 'male'],
+        default=['male', 'male'],
+        help="Speakers' genders for Streaming Text-to-Speech, in `--language_pair` order.")
```
laves:
```suggestion
        help="Gender of the speakers for Streaming Text-to-Speech, in `--language_pair` order.")
```

recipes/live-conversation-translation/python/main.py
```diff
@@ -0,0 +1,260 @@
...
+from pvorca import Orca
+from pvrecorder import PvRecorder
+from pvspeaker import PvSpeaker
+
+
+class LanguagePairs(Enum):
+    DE_EN = "de-en"
+    DE_ES = "de-es"
+    DE_FR = "de-fr"
+    DE_IT = "de-it"
```
laves, line 33: we could derive this by looking at what model files are available in the model folders. Could help with future maintainability

recipes/live-conversation-translation/python/main.py
```diff
@@ -0,0 +1,260 @@
...
+        speaker.start()
+
+        index = 0
+        while True:
+            text = ""
+            text_lock = Lock()
+
+            def get_text() -> str:
+                with text_lock:
+                    return f"HUMAN[{languages[index].upper()}] {text}"
```
Gabe-Picovoice: I originally found it a bit difficult to realize that this demo works in lockstep: it listens for the first language, then translates and speaks out the other one, before alternating. Telling a user it is listening before they speak might be helpful?

```suggestion
            def get_text() -> str:
                with text_lock:
                    display_text = "(listening)" if text == "" else text 
                    return f"HUMAN[{languages[index].upper()}] {display_text}"
```

recipes/live-conversation-translation/python/README.md
```diff
@@ -0,0 +1,65 @@
+# Live Conversation Translation in Python
+
```
laves, line 2: Might be nice to sum up the demo in this readme too (in the case we link to this folder directly and they don't read the other readme)
```
Two-way real-time speech translation powered by on-device AI. Select a pair of languages to translate between and take turns speaking to experience seamless, back-and-forth conversation.
```

recipes/live-conversation-translation/python/setup.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def main() -> None:
+    with open(os.path.join(os.path.dirname(__file__), 'requirements.txt')) as f:
+        for entry in f.read().strip(' \n').split('\n'):
+            if any(x in entry for x in {'cheetah', 'orca', 'zebra'}):
+                package, version = entry.split('~=')
+                animal = package[2:]
+                major, minor = version.split('.')[:2]
+                clone_repo(animal, major, minor)
```
ErisMik, line 84: Some terminal feedback might be nice. Some of these repos are quite large and take a while to clone (even with depth = 1), and its not clear that anything is happening from the script.

Same applies to the other demo.

recipes/live-conversation-translation/python/main.py
```diff
@@ -0,0 +1,260 @@
...
+        '--disable_text_normalization',
+        action='store_true',
+        help='Disable text normalization in Streaming Speech-to-Text.')
+    parser.add_argument(
+        '--genders',
+        nargs=2,
+        choices=['female', 'male'],
+        default=['male', 'male'],
+        help="Speakers' genders for Streaming Text-to-Speech, in `--language_pair` order.")
+    args = parser.parse_args()
```
ErisMik, line 130: An audio-device-index argument would be nice. For example the application does not work by default on my machine because of this and I have to modify the source.

Same applies to the other demo.

recipes/speech-to-speech-translation/python/README.md
```diff
@@ -0,0 +1,73 @@
+# Speech-to-Speech Translation in Python
```
laves, line 1: I think all my comments on the previous demo apply to this one as well.

recipes/live-conversation-translation/README.md
```diff
@@ -0,0 +1,13 @@
+# Live Conversation Translation
+
+Two-way real-time speech translation powered by on-device AI.
```
ErisMik: It wasn't initially obvious to me what the difference between these two translation demos was, maybe some additional description would help?

recipes/live-conversation-translation/python/main.py
kenarsa: That was the first version of the code. It became messy, and I also want people to be able to look at the code and know what is supported.

Yes, keeping this up to date will be work, unfortunately.

recipes/live-conversation-translation/python/README.md
kenarsa: We should move towards this. The newer Ubuntu discourages system-wide package installations to the point that users need to use a special flag to do so.

We don't need to change other docs, but for demos in this repo, let's follow this pattern.

recipes/live-conversation-translation/python/setup.py
kenarsa: We do need all the models for these demos to work. I agree we should give feedback in the terminal, and I changed the code to do so.

recipes/live-conversation-translation/python/main.py
kenarsa: I thought about it, but then we don't do it for the voice LLM demo. I think we should keep it like this for now. If we do it, we need to do it for the speaker as well.

recipes/live-conversation-translation/python/setup.py
laves: I understand they need all the model files, but they don't need our binding code/demo code etc. If we're set on using git to fetch the models, we can checkout just the model files with:
```
git clone --filter=blob:none --depth 1 --sparse git@github.com:picovoice/orca.git
cd orca
git sparse-checkout add lib/common
```

laves (approved):

Gabe-Picovoice (approved):

PR #80, opened 2026-04-14
wake word with speaker verification demos

kenarsa:


recipes/personalized-wake-word/python/enroll.py
```diff
@@ -0,0 +1,133 @@
...
+    porcupine = None
+    recorder = None
+    progress = 0.0
+    animation = None
+
+    try:
+        eagle = pveagle.create_profiler(
+            access_key=access_key,
+            min_enrollment_chunks=eagle_min_enrollment_chunks,
+            voice_threshold=0.)
```
matt200-ok: With a voice threshold of 0.0 if the user waits before speaking a speaker profile of mostly silence will be created. It may be better to allow this as a optional parameter and default it to a low value instead of zero. I found 0.05 worked well when I used the Picovoice wake word.

matt200-ok: Or adding a window length for the enrollment frames could also prevent this excess of silence.

ksyeo1010 (approved):

matt200-ok (approved):

Gabe-Picovoice (approved):

laves (approved):

matt200-ok (approved):

PR #82, opened 2026-04-14
Live conversation translation web

matt200-ok:


laves (commented):
just wanted to give my higher-level feedback before I look at the code. Overall it looks good, just a couple small things:
- The language selection UI was not very clear. Small select box with no label. Let's help users with `Select language pair:` label at least.
- Break the init button to it's own line too
- There should be a way to change the languages once you get to the conversation page besides refresh. Either just a 'back' button that kicks them back to the main page with the access key already populated, or, even better, a way they can change the language while in the conversation page, clearing the text and automatically restarting

recipes/live-conversation-translation/web/package.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "keywords": [
+    "Picovoice",
+    "browser",
+    "voice ai"
+  ],
+  "author": "Picovoice Inc",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@picovoice/cheetah-web": "^4.0.0",
```
ErisMik: Mind updating this to `4.0.1`? It has a few fixes in it.

ErisMik (approved):

recipes/live-conversation-translation/web/src/index.ts
```diff
@@ -0,0 +1,278 @@
...
+};
+
+type PvObject = {
+  audio: AudioStream,
+  cheetah0: CheetahWorker,
+  cheetah1: CheetahWorker,
+  zebra0: ZebraWorker,
+  zebra1: ZebraWorker,
+  orca0: OrcaWorker,
+  orca1: OrcaWorker,
```
laves, line 52: I was finding that I was getting overlapping orca audio outputs. Sometimes the English would play over the German.

laves (commented):
Disregard my comment about en->en being possible, it was because I didn't call `yarn build`. That said, it would have been nice if the app didn't proceed and told me I need to call yarn build if I hadn't

ksyeo1010:
first load starts with de -> de which if you init it gives you an error right away.

ksyeo1010:
image
image


somehow someway, if the first setences is exactly '1 2 3.' with the dot, breaks the app

laves (approved):

laves:
not sure who's going to merge between @ksyeo1010 and @ErisMik, but can one of you give a final check and get this in?

ksyeo1010 (approved):

PR #84, opened 2026-04-16
call screen demo

kenarsa:


recipes/call-screen/python/README.md
```diff
@@ -0,0 +1,71 @@
...
+For example:
+
+```console
+python main.py --access_key ${ACCESS_KEY} --username Alireza
+```
+
+You can also provide `--username_pronunciation` to control how the name is spoken by Orca.
+
+```console
+python main.py --access_key ${ACCESS_KEY} --username Alireza --username_pronunciation AE L IY R EH Z AA
```
Gabe-Picovoice: We should probably mention that the pronunciation is expressed in [ARPAbet](https://en.wikipedia.org/wiki/ARPABET) phonemes, since I wasn't sure about the format until I looked through the orca docs

recipes/call-screen/python/main.py
```diff
@@ -0,0 +1,268 @@
...
+            sys.stdout.write(f"\r{text}{dots}")
+            sys.stdout.flush()
+            residue = len(text) + len(dots)
+
+            i = (i + 1) % len(dots_list)
+            sleep(refresh_sec)
+
+        text = get_text()
+        sys.stdout.write(f"\r{text}    {end}")
+        sys.stdout.flush()
```
matt200-ok: If the input from the user is longer than one line this UI breaks and floods the console with repeated lines. This also happens if the width of the console window is less than the output length of the prompt.

recipes/call-screen/python/main.py
```diff
@@ -0,0 +1,268 @@
...
+            text_thread.join()
+
+            print()
+
+            for x in Actions:
+                print(f"{x.value}. {str(x)}")
+
+            while True:
+                try:
+                    action = Actions(int(input("> ")))
```
laves, line 278: this part I personally found a bit confusing - there was no indication that I needed to enter a number on my keyboard and the actions looked like a list of things I could say. It would be clearer with some instructions, maybe something like

```
Enter a number to select a call-assist action:
[0] Greet
[1] Connect Call
[2] Decline Call
[3] Ask for Details
[4] Ask to Text
[5] Ask to Email
[6] Ask to Call Back
[7] Block Caller
```

recipes/call-screen/python/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+Use `--username` to set the name of the person receiving the call.
+
+For example:
+
+```console
+python main.py --access_key ${ACCESS_KEY} --username Alireza
+```
+
+You can also provide `--username_pronunciation` to control how the name is spoken by Orca.
```
laves: I think we should mention that pronunciation is in ARPABET format otherwise they may be unable to come up with a new pronunciation of their own

Gabe-Picovoice (approved):

matt200-ok (approved):

ErisMik:
It would be nice just to get a small note that says `CALL ENDED` or something like that on the choices that are terminal. Just to make it obvious that was the intended flow and the demo did not crash or something like that.

laves (approved):

PR #86, opened 2026-04-16
Live conversation translation android

matt200-ok:


laves:
Upon using it, I do think we should have a little `src -> dst` language tag above each box

ErisMik (approved):
I think you are missing mentioning Android on the README one level up: https://github.com/Picovoice/pico-cookbook/tree/main/recipes/live-conversation-translation

laves (approved):

PR #87, opened 2026-04-17
speech to speech web

ErisMik:


matt200-ok (approved):

recipes/speech-to-speech-translation/web/setup.py
```diff
@@ -73,15 +85,25 @@ def clone_repo(animal: str, major: str, minor: str) -> None:
...
 
 def main() -> None:
-    with open(os.path.join(os.path.dirname(__file__), 'requirements.txt')) as f:
-        for entry in f.read().strip(' \n').split('\n'):
-            if any(x in entry for x in {'cheetah', 'zebra'}):
-                package, version = entry.split('~=')
-                animal = package[2:]
-                major, minor = version.split('.')[:2]
-                clone_repo(animal, major, minor)
+    public_folder = os.path.join(os.path.dirname(__file__), "public", "models")
```
laves, line 92: we'll need a gitkeep file for this directory - the script errors out if you are using it for the first time

recipes/speech-to-speech-translation/web/src/index.ts
```diff
@@ -0,0 +1,320 @@
...
+  BatWorker
+} from '@picovoice/bat-web';
+import { CheetahTranscript, CheetahWorker } from '@picovoice/cheetah-web';
+import { WebVoiceProcessor } from '@picovoice/web-voice-processor';
+import { ZebraWorker } from '@picovoice/zebra-web';
+import { OrcaWorker } from '@picovoice/orca-web';
+import { AudioStream } from './audio_stream';
+
+const LANGUAGE_PAIRS: any = {
+  automatic: [
```
laves, line 15: I did find the automatic language detection phase a bit awkard. I wasn't sure it was picking up what I was saying at first since it wasn't transcribing anything and it has the dots acting like it was going to transcribe. Then, I just caught the end of it saying it was loading the English model (I had been focused on the transciption dots) and it printed an empty blue bubble followed by the transcription/translation. I think we need to differentiate this step a bit more from transciption

laves (commented):
No need to address this till we're done our Android demos, I am just catching up on reviews

ksyeo1010:
if the audio detected is the same as what we want to translate to, I think it is fine to get rid of the audio heard so far. Also while bat is trying to understand something, I think something visual should be there saying its trying to identify the language

laves (changes requested):
when I selected dst language to en and src to automatic, and I spoke in english, it says in two locations that we can't translate from en to en, but it also locks the UI and I have no way of selecting a new dst language without reloading the demo. I think we need to allow them to go back to language selection.

I also still get a little blank bubble after language detection, followed by the actual translation.

laves (approved):

PR #88, opened 2026-04-22
voice-guided workflow recipes

kenarsa:


recipes/voice-guided-field-reporting/res/context.yml
```diff
@@ -0,0 +1,80 @@
+context:
```
matt200-ok, line 1: If we are going to implement these demo's for multiple platforms these context files should probably live in a shared location

recipes/voice-guided-field-reporting/python/main.py
```diff
@@ -0,0 +1,987 @@
...
+
+class RecipeCompletePromptState(RecipePromptState):
+    def __init__(self, step: OrcaStep) -> None:
+        super().__init__(
+            step=step,
+            prompt="Field report recorded.",
+            next_state=None)
+
+
+def main() -> None:
```
matt200-ok, line 926: The demo does not gracefully catch CTRL-C and if you cancel the demo the cursor does not come back.

recipes/voice-guided-maintenance-and-inspection/res/context.yml
```diff
@@ -0,0 +1,45 @@
...
+    identifyUnit:
+      - "$unitId:unitId"
+
+    reportOilCondition:
+      - "$fluidCondition:fluidCondition"
+      - "(the) oil (level) is $fluidCondition:fluidCondition"
+
+    reportTireCondition:
+      - "$tireCondition:tireCondition"
+      - "(the) tire condition is $tireCondition:tireCondition"
```
matt200-ok: It might be better if condition was optional.

matt200-ok (approved):

PR #93, opened 2026-04-24
call assist recipe 

kenarsa:


ErisMik:
My only comment for the flow is it was not obvious that for the "rhino" section that I had to say exactly one of those phrases. Especially since immediately before the input is "free form" from the call. 

Maybe changing the prompt for the section? And/or wrapping the options in quotes so they hint towards more literal?

matt200-ok (commented):
I found that the llm had a hard time determining both who was calling and the reason. It might be better to allow the program to remember the fields so if the llm doesn't catch the reason you don't need to repeat who is calling.

recipes/call-assist/python/main.py
```diff
@@ -0,0 +1,471 @@
...
+                    print(
+                        f"[AI] Unknown caller is trying to speak with you about `{reason}`. "
+                        f"I will ask for their identity.")
+                else:
+                    print(f"[AI] `{caller}` is trying to speak with you. I will ask for their reason.")
+
+                if ask_for_details_retry_count < ask_for_details_retry_limit:
+                    action = Actions.ASK_FOR_DETAILS
+                else:
+                    action = Actions.DECLINE_CALL
```
Gabe-Picovoice: It should probably not print "I will ask for more information." before declining the call.


```suggestion
                if ask_for_details_retry_count < ask_for_details_retry_limit:
                    action = Actions.ASK_FOR_DETAILS
                    if caller == 'unknown' and reason == 'unknown':
                        print("[AI] Unknown caller with no specific reason. I will ask for more information.")
                    elif caller == 'unknown':
                        print(
                            f"[AI] Unknown caller is trying to speak with you about `{reason}`. "
                            f"I will ask for their identity.")
                    else:
                        print(f"[AI] `{caller}` is trying to speak with you. I will ask for their reason.")
                else:
                    action = Actions.DECLINE_CALL
                    print("[AI] Unknown caller has exceeded their retry limit. Declining their call.")
                


```

recipes/call-assist/python/main.py
```diff
@@ -0,0 +1,471 @@
...
+                print(f"[AI] `{caller}` is trying to speak with you about `{reason}`.")
+                break
+
+            ask_for_details_retry_count += 1
+
+        synthesize_and_playback(
+            orca=orca,
+            speaker=speaker,
+            text=f"{caller} is trying to speak to you about {reason}.")
+
```
Gabe-Picovoice, line 411: We should probably make the above a function returning text that relates to the action being performed. If the caller is denied, it should either say that, or even say nothing for the brevity of the user, rather than "unknown is trying to speak to you about unknown"

recipes/call-assist/python/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+1. Open [Picovoice Console](https://console.picovoice.ai/)
+2. Go to Rhino Speech-to-Intent.
+3. Create an empty Rhino context.
+4. Click Import YAML in the top-right corner.
+5. Paste the [Rhino context YAML](../res/context.yml) for this demo.
+6. Download the generated Rhino context file (`.rhn`) for your target platform.
+
+### 6. Run the Demo
+
```
Gabe-Picovoice, line 72: Brief instructions on how to interact with the demo would be helpful for demo users, and for users adapting the code.

```suggestion
### 6. Run the Demo

This demo simulates responding to an incoming phone call using the current device's microphone. Start by speaking as the caller, then receive guidance from the assistant on how to respond as the user.

```

Gabe-Picovoice (commented):
Small issue with the wrong text being displayed when passing the retry count.

recipes/call-assist/python/main.py
kenarsa: It can't reach this point unless both the caller and the reason are known.

Gabe-Picovoice: ah, good point. I missed the loop above. Looks good then!

Gabe-Picovoice (approved):

kenarsa:
@matt200-ok, it makes the demo process too complex. I think for a demo, I would rather keep one system prompt and no-chat-memory LLM.

matt200-ok (approved):

recipes/call-assist/python/main.py
```diff
@@ -0,0 +1,473 @@
...
+    def prompt(self, username: str) -> str:
+        return {
+            self.GREET: "Hi, {username} can't answer right now. Please say your name and why you're calling.",
+            self.CONNECT_CALL: "Okay, one moment while I connect you.",
+            self.DECLINE_CALL: "Sorry, {username} is unavailable right now.",
+            self.ASK_FOR_DETAILS: "Can you briefly say who you are and what this is regarding?",
+            self.ASK_TO_TEXT: "{username} can't talk right now. Please send a text message instead.",
+            self.ASK_TO_EMAIL: "Please send the details by email. Thank you.",
+            self.ASK_TO_CALL_BACK: "{username} can't take your call right now. Please call back later.",
```
laves: super nick-picky, but when I don't set a username `ASK_TO_TEXT` and `ASK_TO_CALL_BACK` sentences start with `the receiptient`, it should start with `The receiptient` with the capitol at the beginning

laves (approved):

laves (approved):

ErisMik (approved):

ksyeo1010 (approved):

PR #98, opened 2026-04-29
Personalized Wake Word for Android

laves:


recipes/personalized-wake-word/android/personalized-wake-word/src/main/java/ai/picovoice/personalizedwakeword/MainActivity.java
```diff
@@ -0,0 +1,353 @@
...
+                Log.e(TAG, "Audio processing error: ", e);
+            }
+        };
+
+        errorListener = error -> Log.e(TAG, "Audio Error: ", error);
+        voiceProcessor.addFrameListener(frameListener);
+        voiceProcessor.addErrorListener(errorListener);
+    }
+
+    private void checkPermissionsAndStart(AppState targetState) {
```
matt200-ok, line 169: In the other demos we also start when permission is granted and I think you should do the same here.

recipes/personalized-wake-word/android/personalized-wake-word/src/main/java/ai/picovoice/personalizedwakeword/MainActivity.java
```diff
@@ -0,0 +1,353 @@
...
+        if (!voiceProcessor.hasRecordAudioPermission(this)) {
+            ActivityCompat.requestPermissions(this, new String[]{Manifest.permission.RECORD_AUDIO}, 1);
```
matt200-ok: The app should produce an error message when microphone permissions are not granted. It currently just gets stuck with the button not doing anything.

matt200-ok (commented):
It would also be good to add a gitkeep to the assets folder so the user does not need to create it.

matt200-ok (approved):

ErisMik (approved):

PR #100, opened 2026-04-29
Call Screen Demo [Android]

Gabe-Picovoice:
I also made some small fixes to the other related demos.


Gabe-Picovoice:
I'm not sure how much sharing we want in terms of visual style between demos, but for polish I included:
- animated dots when listening
- coloured text for when users speak
- fade between scenes
- demo can be replayed

recipes/call-screen/android/README.md
```diff
@@ -0,0 +1,37 @@
...
+3. Type your name in the `USERNAME` variable in [MainActivity.java](call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java).
+
+### 2. Train the Speech-to-Intent Model
+
+1. Open [Picovoice Console](https://console.picovoice.ai/)
+2. Go to Rhino Speech-to-Intent.
+3. Create an empty Rhino context.
+4. Click Import YAML in the top-right corner.
+5. Paste the [Rhino context YAML](../res/context.yml) for this demo.
+6. Select the **Android** platform and download the generated Rhino context file (`.rhn`).
```
laves, line 25: need a step for adding the rhn file to the project asset folder

recipes/call-screen/android/call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java
```diff
@@ -0,0 +1,607 @@
...
+        LISTEN_TO_USER
+    }
+
+    private static final String ACCESS_KEY = "${YOUR_ACCESS_KEY_HERE}";
+    private static final String USERNAME = "${YOUR_USERNAME_HERE}";
+
+    private static final String STT_MODEL_FILE = "cheetah_params.pv";
+    private static final String TTS_MODEL_FILE = "orca_params_female.pv";
+    // TODO: add instructions that this file must be generated then moved here
+    private static final String RHINO_CONTEXT_FILE = "call-screen-demo.rhn";
```
laves: yes, we need to add this to the readme. Also add .rhn to gitignore for this project

recipes/call-screen/android/call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java
```diff
@@ -0,0 +1,607 @@
...
+
+    private static final String PUNCTUATION = "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~";
+
+    private static final int TTS_WARMUP_SECONDS = 1;
```
laves: unused?

recipes/call-screen/android/call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java
```diff
@@ -0,0 +1,607 @@
...
+
+    @Override
+    public void onRequestPermissionsResult(
+        int requestCode,
+        @NonNull String[] permissions,
+        @NonNull int[] grantResults
+    ) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults);
+        if (grantResults.length == 0 || grantResults[0] == PackageManager.PERMISSION_DENIED) {
+            onEngineProcessError("Recording permission not granted");
```
laves: I prefer printing `Recording permission is required for this demo` and then still giving them the button to start. That way then can try again and accept

recipes/call-screen/android/call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java
```diff
@@ -0,0 +1,607 @@
...
+
+    private void enableStartButton() {
+        voiceProcessor.addFrameListener(this::frameListener);
+        voiceProcessor.addErrorListener(error -> {
+            onEngineProcessError(error.getMessage());
+        });
+
+        mainHandler.post(() -> loadingText.setText("Press the Start Demo button to begin."));
+        startButton.setOnClickListener(view -> {
+            // TODO: fix double click; this is not working
```
laves: please address all TODOs and remove from code before submitting for review

recipes/call-screen/android/call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java
```diff
@@ -0,0 +1,607 @@
...
+                        animation.end();
+                        animation = null;
+
+                        new Thread(() -> {
+                            speakToCaller(action);
+                        }).start();
+                    } else {
+                        appendStyledText(
+                                userTextBuilder,
+                                "Unknown Action\n" + "[" + username.toUpperCase() + "] ",
```
laves: I think it may look better if we keep the listening state rather than printing `Unknown Action` and starting a new listening

recipes/call-screen/python/main.py
```diff
@@ -209,7 +233,11 @@ def main() -> None:
             require_endpoint=False)
         print(f"[OK] Rhino Speech-to-Intent[V{rhino.version}]")
 
-        recorder = PvRecorder(frame_length=cheetah.frame_length)
+        recorder = PvRecorder(
+            device_index=input_device_index,
+            frame_length=cheetah.frame_length)
+        print(f"[OK] PV Recorder[V{recorder.version}]")
+        print(f"     Recording Device {recorder.selected_device}")
```
laves: I understand why you want to do this, but let's maybe leave this out for now. I think we'd want to add to all demos if we want to start printing pv recorder stuff along with the engine feedback

Gabe-Picovoice: Oh, do you mean just these two prints, or the ability to configure the mic?

laves: I did just mean the prints, though we also will need to add mic configuration to all demos, so maybe the whole change should just be avoided for now. Why don't you create a card to do it and we'll do it in a coordinated change

Gabe-Picovoice: sounds good! I was thinking your logic applied to everything

laves:
btw Eric had an interesting solution to the problem of having to store cheetah and orca models, and copy the rhino model over. Following some of the other cookbook demos a [python script was used](https://github.com/Picovoice/pico-cookbook/pull/101/changes#diff-a2e107a661152fd1e1301f617c6b4a1e9513ed1c99f07a9b01ba9d362f3a0102)

recipes/call-screen/android/call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java
Gabe-Picovoice: Good idea! I made it so that "Unknown Action" only appears for 1000ms and is ephemeral, so it works as a notice, but doesn't take up any state.

Gabe-Picovoice:
> btw Eric had an interesting solution to the problem of having to store cheetah and orca models, and copy the rhino model over. Following some of the other cookbook demos a [python script was used](https://github.com/Picovoice/pico-cookbook/pull/101/changes#diff-a2e107a661152fd1e1301f617c6b4a1e9513ed1c99f07a9b01ba9d362f3a0102)

Cool! I'll try his approach

matt200-ok (commented):
* I think when the demo finishes you should have a longer delay before going back to the home screen.
* I think instead of putting the call assist actions in the second box I think you should display it as a tool tip like the android voice memo demo.
* I think it might be better to remove the second box entirely and put all the interactions in the first box.

Gabe-Picovoice:
> * I think when the demo finishes you should have a longer delay before going back to the home screen.
> 
> * I think instead of putting the call assist actions in the second box I think you should display it as a tool tip like the android voice memo demo.
> 
> * I think it might be better to remove the second box entirely and put all the interactions in the first box.

I tried out the voice memo demo and I think I see what you mean. I can replace the second box with a textview and display what command the user selected by changing the style of the text. After the command is selected, I can hide the tooltip.

I think putting *all* the interactions in the first box would make it look like the caller and the user are speaking to each other, despite that the point of demo is that the screening separates the two.

Gabe-Picovoice:
I kept the action selection logic pretty similar, but now it appears more like a tooltip that fades away after being selected.

I also added a small [SYSTEM] message so it's more clear to the user what the demo is simulating.

I think this is less overwhelming at first, since the user doesn't have to think about why there are two text boxes.

matt200-ok (commented):
On my phone the tool tip overlaps with other text. You should only print necessary information in the tool tip, reduce the font size slightly and make sure everything scrolls correctly.

recipes/call-screen/android/README.md
```diff
@@ -0,0 +1,56 @@
...
+your AccessKey with Picovoice license servers, even though the inference is running 100% offline. Everyone who signs up
+for [Picovoice Console](https://console.picovoice.ai/) receives a unique AccessKey.
+
+## Usage
+
+### 1. Configure the demo
+
+1. Open the `CallScreen` project in Android Studio. 
+2. Copy your `AccessKey` from Picovoice Console into the `ACCESS_KEY` variable in [MainActivity.java](call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java).
+3. Type your name in the `USERNAME` variable in [MainActivity.java](call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java).
```
laves: this is handled with a setup script now right?

recipes/call-screen/android/call-screen/src/main/res/layout/main_layout.xml
```diff
@@ -0,0 +1,186 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:visibility="visible">
+
+        <androidx.constraintlayout.widget.Guideline
+            android:id="@+id/callerGuidelineHorizontal"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:orientation="horizontal"
+            app:layout_constraintGuide_percent="0.50" />
```
laves: I do find the chat layout maybe a little too large and the call assist suggestions area ends up looking a bit crunched in comparison. I think you can probably give the chat box 10% less and the suggestion area 10% more screen

recipes/call-screen/android/call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java
```diff
@@ -0,0 +1,683 @@
...
+                    break;
+                case SPEAKING_TO_CALLER:
+                    stateText.setText("AI: Speaking to caller");
+                    break;
+                case LISTEN_TO_USER:
+                    stateText.setText("AI: Listening for " + username + "'s command");
+                    break;
+                default:
+                    stateText.setText("AI: Unknown state");
+                    break;
```
laves: probably don't want to admit we got in a weird state in the UI. If anything throw an error

recipes/call-screen/android/README.md
Gabe-Picovoice: thanks for catching this!

Gabe-Picovoice:
I tested UI on both medium phone and small phone. Not sure why small phone degrades performance pretty heavily on my system, but I can confirm the text is visible & the scroll works.

matt200-ok (approved):

recipes/call-screen/android/README.md
```diff
@@ -15,11 +15,7 @@ for [Picovoice Console](https://console.picovoice.ai/) receives a unique AccessK
 
 ## Usage
 
-### 1. Configure the demo
-
-1. Open the `CallScreen` project in Android Studio. 
-2. Copy your `AccessKey` from Picovoice Console into the `ACCESS_KEY` variable in [MainActivity.java](call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java).
-3. Type your name in the `USERNAME` variable in [MainActivity.java](call-screen/src/main/java/ai/picovoice/callscreen/MainActivity.java).
+### 1. Open the `CallScreen` project in Android Studio
```
laves: I think you can remove this usage step and just cover that in the `3. Run the Android Studio project`. Just make sure you update the step numbers below accordingly

Gabe-Picovoice: ah, nice catch!

laves (approved):

PR #101, opened 2026-04-29
Initial android voice memo demo

ErisMik:


recipes/voice-memo-assistant/android/setup.py
```diff
@@ -0,0 +1,132 @@
...
+        thread.join()
+
+    sys.stdout.write(f"\rCloned {animal.capitalize()}'s GitHub repository.{' ' * 20}\n")
+    sys.stdout.flush()
+
+    return folder
+
+
+def main() -> None:
+    parser = ArgumentParser()
```
laves, line 96: maybe this is a bad idea, but what if we accept `--access_key` as an argument and get it to do the find and replace in `MainActivity.java`. Then it sort of becomes and all-in-one setup script

recipes/voice-memo-assistant/android/settings.gradle
```diff
@@ -0,0 +1,2 @@
+rootProject.name = "LLMVoiceAssistant"
```
laves:
```suggestion
rootProject.name = "VoiceMemoAssistant"
```

recipes/voice-memo-assistant/android/README.md
```diff
@@ -0,0 +1,57 @@
+# Voice Memo Assistant in Android
+
```
laves, line 2: I see the original demo didn't include a description, so we don't have one here. Could you add a short description to `recipes/voice-memo-assistant/README.md` and this one?

matt200-ok (commented):
When asking to read an empty memo such as when saying start recording then only saying stop recording. Orca errors out with Text cannot be pronounced.

ErisMik:
> When asking to read an empty memo such as when saying start recording then only saying stop recording. Orca errors out with Text cannot be pronounced.

Fixed!

matt200-ok (approved):

recipes/voice-memo-assistant/android/setup.py
ErisMik: I kinda like it, made the change!

recipes/voice-memo-assistant/android/README.md
ErisMik: Actually there was one in the other READMEs now (This branch was out of date). Updated Android to match them.

laves (approved):

PR #102, opened 2026-04-29
Speech to speech translation android

matt200-ok:


laves:
I found the automatic language detection a little awkward because here's no indication I should say anything to help it detect the source language. I think at very least we should prompt the user to 'Speak in the source language...`. Labels or icons or something to help us understand that the left spinner is for souce language and the right is for destination would also be nice

laves:
- when I init the demo without AccessKey and automatic language detection, said it was detecting the source language, but with red text
- when I start the demo, the mic button is a pause button. This doesn't really make sense because it isn't listening for anything yet. We should probably do grey mic until languages have been selected. It's also sort of weird that the mic button is usable while the models are being loaded. I think it's more convetional to lock user interaction in that case

laves (approved):

PR #103, opened 2026-04-29
vlm and ocr recipes

kenarsa:


recipes/image-qa/python/main.py
```diff
@@ -0,0 +1,420 @@
...
+        tts_thread.join()
+
+        if not tts_error_queue.empty():
+            raise tts_error_queue.get()
+
+        answer_event.set()
+        answer_thread.join()
+
+        with answer_lock:
+            return answer.strip()
```
Gabe-Picovoice, line 308: Lock no longer needed here since `answer_thread` is joined and `generate_with_image` has returned
```suggestion
        return answer.strip()
```

recipes/image-qa/python/main.py
```diff
@@ -0,0 +1,420 @@
...
+    progress_lock = Lock()
+
+    sentence_end_pattern = re.compile(r"([.!?;:])\s+")
+
+    def get_answer() -> str:
+        with answer_lock:
+            if len(answer) == 0:
+                with progress_lock:
+                    if len(progress) > 0:
+                        return progress
```
Gabe-Picovoice, line 152: OCR also has this function.

```suggestion
                        return f"[A] (thinking {progress})"
```

ErisMik:
While it's processing the image or outputting tokens, I can't seem to stop the demo. The ctrl-c gets captured and spits an error but the demo persists. (for image-qa demo)

matt200-ok (commented):
When using the image-to-speech demo the model sometimes outputs markdown which doesn't do well when passed into orca.

kenarsa:
@matt200-ok it is what it is for now. @ErisMik I think that is a side effect of picollm.

recipes/image-qa/python/main.py
```diff
@@ -0,0 +1,420 @@
...
+                return "[A] (thinking)"
+
+            return f"[A] {answer}"
+
+    def update_progress(x: float) -> None:
+        nonlocal progress
+        with progress_lock:
+            progress = f"{x:.2f}%"
```
laves: Could be nice to contextualize what this number is for with something like `Processing image (##.#%)...`  or `Processing question (##.#%)...` or something

recipes/image-to-speech/python/main.py
```diff
@@ -0,0 +1,368 @@
...
+                    if len(progress) > 0:
+                        return progress
+                return "[OCR] (reading image)"
+
+            return f"[OCR] {extracted_text}"
+
+    def update_progress(x: float) -> None:
+        nonlocal progress
+        with progress_lock:
+            progress = f"{x:.2f}%"
```
laves: This one as well would be nice if it had some descriptive text to indicate what is happening

matt200-ok (approved):

laves (commented):
related to @ErisMik comment on the ctrl-C not doing anything, when I've waited for picoLLM to finish it's output and I just want to stop Orca from reading the long output, I hit ctrl-C and it actually segfaulted in both demos. Not sure if that's an us issue or a python issue depending on what ctrl C connects to

kenarsa:
@laves should be addressed.

laves (approved):

PR #105, opened 2026-04-30
Document QA Android

ErisMik:


matt200-ok (approved):

recipes/document-qa/android/document-qa/src/main/java/ai/picovoice/documentqa/MainActivity.java
```diff
@@ -0,0 +1,1025 @@
...
+                    }
+
+                    engineExecutor.submit(() -> {
+                        initEngines(selectedUri);
+                    });
+                }
+            });
+
+    private ArrayList<String> chunkDocument(Uri uri) {
+        // TODO: Sub and strip newlines?
```
laves: still TODO?

laves (approved):

recipes/document-qa/android/document-qa/src/main/java/ai/picovoice/documentqa/MainActivity.java
ErisMik: No longer applicable :+1:

recipes/document-qa/android/README.md
```diff
@@ -0,0 +1,56 @@
...
+Copy your downloaded picoLLM model files into the assets folder, located at `/src/main/assets`.
+
+### 3. Add AccessKey to MainActivity.java
+
+Open the project in Android Studio and open `MainActivity.java`. Change the value of the `ACCESS_KEY` variable to your Picovoice AccessKey.
+
+### 4. Run the Demo with Android Studio
+
+With an Android device connected or an Android simulator running, build and run the demo using Android Studio.
+
```
Gabe-Picovoice, line 56: Is there a reason you're not using the `setup.py` solution for this demo (downloading cheetah params + providing access_key)? Ian recommended I use that approach after seeing yours.

recipes/document-qa/android/document-qa/src/main/res/drawable/button_secondary.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Gabe-Picovoice: is this file being used? Might be a good idea to clear some of these up if they're unused.

Gabe-Picovoice (commented):
Looks good and works well (especially the cache)! Although I noticed it can be tricky to ask the right questions if you pass in a different kind of document than what it's looking for.

recipes/document-qa/android/README.md
ErisMik: Just that there are less model files for the user to deal with this one vs. the other one so tried to keep it simple, but if you feel like the `setup.py` would help here I can add it.

ErisMik:
> Looks good and works well (especially the cache)! Although I noticed it can be tricky to ask the right questions if you pass in a different kind of document than what it's looking for.

Yeah unfortunately I had to tune down the context to make the demo responsive as the phone isn't as powerful as the other platforms, so the responses aren't as helpful.

Gabe-Picovoice:
But they do show the demo is working at least, so it's understandable but okay!

recipes/document-qa/android/README.md
Gabe-Picovoice: I think it's fine without it. setup.py is nice, but it takes a while to clone.

Gabe-Picovoice (approved):

PR #110, opened 2026-05-04
Personalized Wake Word demo for Web

laves:


matt200-ok (commented):
You should add a check to see if Picovoice exists so calling `yarn start` without calling `yarn build` doesn't brick the system. Also it might be better to always overwrite the keyword file.

ErisMik (approved):

ErisMik:
Looks great!

matt200-ok (approved):

PR #112, opened 2026-05-05
Call Assist iOS

matt200-ok:


recipes/call-assist/ios/README.md
```diff
@@ -0,0 +1,67 @@
...
+'connecting the call, declining it, asking for more details, requesting a text or email, asking the caller to call back
+later, or blocking the caller.
+
+The entire interaction is powered by on-device AI, combining real-time speech-to-text, text-to-speech, speech-to-intent,
+and local language model inference to create a private, low-latency call-screening experience without sending audio or
+call content to the cloud.
+
+## Compatibility
+
+- iOS 16.0+
```
laves: When I loaded the project, it didn't have an `IPHONEOS_DEPLOYMENT_TARGET` defined so I couldn't launch it on my simulator. I added it, with target = iOS 16, but one of the UI components requires 17+

laves (commented):
Some UI thoughts:

There's a lot of dead space in the demo. When I ran it on an iPhone 16 simulator, most of the demo happened in the upper 1/3 of the screen. I think we should make the layout fill out a bit more of the UI. Also I didn't see a VU meter, which the Android demo has an I think helps the demo look more filled out and sophisticated.

ErisMik:
Just need to fix the `info.plist` issue to run on an actual device, but other than that looks good to me!

ErisMik (approved):

laves (approved):

PR #118, opened 2026-05-06
Voice Memo demo for Web

laves:


recipes/voice-memo-assistant/web/src/main.ts
```diff
@@ -0,0 +1,245 @@
...
+const handleIntent = async (intent: string | undefined) => {
+  if (intent === 'startMemo') {
+    memoText = '';
+    enhancedText = '';
+    setState(UIState.START_RECORDING);
+    await WebVoiceProcessor.subscribe(object!.cheetah);
+  } else if (intent === 'readMemo') {
+    setState(UIState.READ_RECORDING);
+    const textToRead = enhancedText || NO_MEMO_ERROR_PHRASE;
+    playAudioMessage(textToRead);
```
matt200-ok, line 171: The UI will always flash reading memo even if there is no memo recorded for a moment.

matt200-ok (commented):
I think it would be better to center the demo more like the llm voice assistant.

matt200-ok (approved):

recipes/voice-memo-assistant/web/.gitignore
ErisMik, line 1: The cloned setup repos aren't git-ignored

ErisMik (commented):
When the demo is reading the memo it goes back to the wake word state before the audio is finished playing. I'm not sure you can actually trigger the wake word but it's showing the prompt. If it's not too hard we should probably stay in the "reading" state prompt until the audio is finished playing?

ErisMik (approved):

PR #122, opened 2026-05-08
Document QA iOS

matt200-ok:


laves:
- ensure caching is working
- tooltip or instructions would be helpful
- add VU Meter below text box to show user when we want them to say something - change to rotating progress circle while llm is running

recipes/document-qa/ios/DocumentQA/MainView.swift
```diff
@@ -72,6 +72,20 @@ struct MainView: View {
                     RoundedRectangle(cornerRadius: 10)
                         .stroke(Color.lightGray, lineWidth: 1)
                 )
+
+            Text(viewModel.listenState == .question ? "Ask a Question" : " ")
```
laves: It would be nice to have a `Thinking...` and a `Responding...` text above the progress spinner rather than just no text at all

PR #133, opened 2026-05-13
Call Screen Demo [Web]

Gabe-Picovoice:
- restartable
- VU meter
- objects transition instead of appearing abruptly


recipes/call-screen/web/setup.py
```diff
@@ -0,0 +1,122 @@
...
+
+def main() -> None:
+    parser = ArgumentParser()
+    parser.add_argument(
+        '--context_path',
+        required=True,
+        help='Absolute path to the Rhino context file (`.rhn`).')
+    args = parser.parse_args()
+
+    public_folder = os.path.join(os.path.dirname(__file__), "public", "models")
```
matt200-ok, line 106: You need to add a .gitkeep to the models directory

recipes/call-screen/web/public/index.html
```diff
@@ -0,0 +1,76 @@
...
+        </label>
+        <input type="text" id="access-key" name="access-key"/>
+      </div>
+
+      <div id="name-container">
+        <label for="name">Your name</label>
+        <input type="text" id="name" name="name" />
+      </div>
+
+      <button id="init" disabled>
```
matt200-ok: I think you should make the sign in screen not visible when the demo is running and make the demo screen not visible when you are signing in. Also after inserting an access key once it should never be visible again so I think that field should disappear once the engines have been loaded. Also The engines should also only be loaded once.

Gabe-Picovoice: Oh, I see what you mean. I'll separate inputting the AccessKey + loading from the rest of the demo in a single interaction. Then the rest of the demo will allow you to select a name & start, in addition to changing the name between iterations.

recipes/call-screen/web/setup.py
Gabe-Picovoice: ty!

matt200-ok (approved):

recipes/call-screen/web/setup.py
```diff
@@ -0,0 +1,122 @@
...
+
+    shutil.copy(args.context_path, os.path.join(public_folder, "call_screen_demo_web.rhn"))
+    print(f"Copied ${args.context_path} to public/models/call_screen_demo_web.rhn")
```
ErisMik: Python not JS ;)

```suggestion
    print(f"Copied {args.context_path} to public/models/call_screen_demo_web.rhn")
```

Gabe-Picovoice: ty!

ErisMik (approved):

ErisMik:
Oh sorry one last thing, you missed adding `web` to the recipe readme: https://github.com/Picovoice/pico-cookbook/blob/main/recipes/call-screen/README.md

Gabe-Picovoice:
> Oh sorry one last thing, you missed adding web to the recipe readme

Oh right, good catch!

PR #134, opened 2026-05-13
Live Captioning & Translation Web

ErisMik:


recipes/live-captioning-and-translation/web/public/index.html
```diff
@@ -0,0 +1,176 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <script src="index.js"></script>
```
matt200-ok, line 4: The demo should show an error message when index.js does not exists telling the user to do a yarn build

recipes/live-captioning-and-translation/web/public/index.html
```diff
@@ -0,0 +1,176 @@
...
+      </div>
+
+      <div id="message" class="messageArea">
+        <div id="meterContainer" style="display: none">
+          <div id="bar1" class="volume-bar"></div>
+          <div id="bar2" class="volume-bar"></div>
+          <div id="bar3" class="volume-bar"></div>
+        </div>
+        <div id="resetContainer">
+          <button id="reset">Reset demo</button>
```
matt200-ok, line 180: During the mic demo the reset button gets pushed off the screen in my browser

matt200-ok, line 180: Also when hitting reset while the file is speaking I got a crash where cheetah was being used after being deleted.

matt200-ok (approved):

PR #136, opened 2026-05-13
Call Assist Demo [Web]

Gabe-Picovoice:
- Spinner while LLM is processing
- Model is not very good, so logic to handle bad responses better


ErisMik (approved):

recipes/call-assist/web/setup.py
```diff
@@ -0,0 +1,131 @@
...
+import time
+from argparse import ArgumentParser
+from threading import (
+    Event,
+    Thread
+)
+from typing import Tuple
+
+ANIMALS = [
+    ("cheetah", "4.0.0"),
```
laves: 4.0.3 is latest

recipes/call-assist/web/src/index.ts
```diff
@@ -0,0 +1,423 @@
...
+
+  // If you update your model, you may want to enable force write to update the model in storage.
+  const FORCE_WRITE = false;
+
+  sendMessage("status", "Loading Cheetah");
+  const cheetah = await CheetahWorker.create(
+    accessKey,
+    transcriptCallback,
+    { publicPath: "models/cheetah_params.pv", forceWrite: FORCE_WRITE },
+    {
```
laves: for the demo, just enable it. If folks want to learn more they can read the def in the API docs
```suggestion
  sendMessage("status", "Loading Cheetah");
  const cheetah = await CheetahWorker.create(
    accessKey,
    transcriptCallback,
    { publicPath: "models/cheetah_params.pv", forceWrite: true },
    {
```

recipes/call-assist/web/src/index.ts
```diff
@@ -0,0 +1,423 @@
...
+        if (object !== null && rhinoPcmBuffer.length >= object!.rhino.frameLength) {
+          object!.rhino.process(rhinoPcmBuffer.slice(0, object!.rhino.frameLength));
+          rhinoPcmBuffer = rhinoPcmBuffer.slice(object!.rhino.frameLength)
+        }
+        break;
+    }
+  }
+}
+
+// ------------------------------------------------------------------------- //
```
laves: can you remove these dividers? they're not really consistent with our style and don't add much imho. This applies in other files as well

recipes/call-assist/web/src/index.ts
```diff
@@ -0,0 +1,423 @@
...
+    const normalized = (db - MIN_DB) / (MAX_DB - MIN_DB);
+    const normalizedVolume = Math.max(0.0, Math.min(1.0, normalized));
+
+    onVolumeCallback(normalizedVolume);
+  }
+
+  const resetDemo = async () => {
+    object!.audio.clear();
+
+    sendMessage("ai state", "");
```
laves: even if they present as strings we should codify they message types so that we know what's available, e.g.:
```typescript
let messageTypes: 'IDLE' | 'AI_STATE' | 'RESTART_DEMO' | 'NEW_AI_BUBBLE' = 'IDLE';
```

recipes/call-assist/web/src/index.ts
```diff
@@ -0,0 +1,423 @@
...
+    const inference = completion.completion.trim().replace('<|eot_id|>', '');
+    console.log(`# Inference\n${inference}`);
+    const [caller, reason] = extractCallerAndReasonFromLLMInference(inference);
+
+    sendMessage("stop llm spinner", null);
+
+    if (caller != 'unknown' && reason != 'unknown') {
+      sendMessage(
+          "ai report",
+          `[AI] <b style="color: var(--brand-primary);">${caller}</b> is trying to speak with you about ` +
```
laves: the inline text could be a util function:

```typescript
const styleKeyText = (text: string) => `<b style="color: var(--brand-primary);">${text}</b>`;
```

recipes/call-assist/web/src/index.ts
```diff
@@ -0,0 +1,423 @@
...
+      } else if (caller == 'unknown') {
+        sendMessage(
+            "ai report",
+            `[AI] Unknown caller is trying to speak with you about <b style="color: var(--brand-primary);">${reason}</b>. ` +
+            `I will ask for their identity.`);
+      } else {
+        sendMessage(
+            "ai report",
+            `[AI] <b style="color: var(--brand-primary);">${caller}</b> is trying to speak with you. I will ask for their reason.`);
+      }
```
laves, line 307: you could probably simplify the logic by either setting caller name to `${caller}` or `Unknown caller` whether we know them or not, e.g.:
```typescript
let callerName = caller == 'unknown' ? 'Unknown caller' : styleKeyText(caller);
```

laves (changes requested):
- more centralized home screen layout (can base off personalized wake word)
- name and start demo button can go away and become part of the home screen, demo can start automatically
- VU meter at bottom of phone layout instead of moving along with text bubbles
- ideally have one central spot for status updates
- response options need a different style so that we don't think they're buttons

recipes/call-assist/web/src/index.ts
Gabe-Picovoice: Great idea! I added a type for this

recipes/call-assist/web/setup.py
Gabe-Picovoice: good catch, ty!

recipes/call-assist/web/src/index.ts
Gabe-Picovoice: great suggestion!

recipes/call-assist/web/src/index.ts
Gabe-Picovoice: I updated this segment to use this pattern & it's more readable now, but the logic wasn't simplified too much unfortunately.

laves (approved):

PR #137, opened 2026-05-14
Live Captioning and Translation iOS

matt200-ok:


ErisMik (approved):

laves:
Few UI notes:
- it's really not clear what type of file is supported when I click the Use File button
- it may be nice to have those three dots going in the text box before audio has been detected. The empty box just looks kind of sad, and we've set the convention of having those dots in other demos
- cut new bubble once you have translation
- I wasn't able to test it on a real iPhone because the one I have right now is iOS 16 and this demo is 17+, so be sure you test it on a real device. I also wasn't able to try a file because of my simulator not accepting files for some reason.
- that home page looks a little barren in the middle. Maybe we put a title in our brand colour just to give the sense that this is the demo 'home screen'

laves (commented):
I just got a crash when I selected `Use File`, but then cancelled the file picking. Also can we say `Use Audio File` to make it dead obvious what kind of file we're accepting

laves (commented):
I just got a crash when I selected `Use File`, but then cancelled the file picking. Also can we say `Use Audio File` to make it dead obvious what kind of file we're accepting

PR #138, opened 2026-05-15
Live Captioning and Translation Android

matt200-ok:


ErisMik (commented):
Missing recipe level adding Android to README :)

ErisMik (approved):

laves (commented):
- the error message was cut-off on the left of the screen when I launched the demo without replacing AccessKey
- when loading models, the status text is at the top, which looks a bit odd to me. I believe in the iOS demo is was over the language selection boxes, which makes more sense to me
- really minor thing, but when I didn't accept mic permissions, then went back to the home screen, the status text was red
- it should be really obvious and documented that only 16k, mono WAV files are accepted and if the user gives us anything else we should be clear what the issue is with the file. Or if we can accept different formats, great, but we need them to work properly

recipes/live-captioning-and-translation/android/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### 2. Build and Run the Demo
+
+1. Open the `live-captioning-and-translation` project in Android Studio.
+
+2. Copy your `AccessKey` from Picovoice Console into the `ACCESS_KEY` variable in [MainActivity.java](live-captioning-and-translation/src/main/java/ai/picovoice/livecaptioningandtranslation/MainActivity.java).
+
+3. Connect a device or launch an Android simulator.
+
+4. Press the `Use Mic` or `Use Audio File` button. The demo only supports 16khz single channel, 16 bit audio files.
```
laves:
```suggestion
4. Press the `Use Mic` or `Use Audio File` button. The demo only supports 16kHz single-channel, 16-bit WAV files.
```

recipes/live-captioning-and-translation/android/setup.py
```diff
@@ -0,0 +1,113 @@
...
+import time
+from threading import (
+    Event,
+    Thread
+)
+from typing import Tuple
+
+
+ANIMALS = [
+    ("cheetah", "4.0.1"),
```
laves: use latest cheetah

laves (approved):

PR #141, opened 2026-05-19
Voice Guided Maintenance and Inspection Android

matt200-ok:


ErisMik (commented):
Only comment:
- Title should be text align centered so it looks less awkward when broken into multiple lines

laves (approved):

PR #142, opened 2026-05-19
Call Screen Demo Patch #3 [Android]

Gabe-Picovoice:
Ensure `userTextBuilder` modifications only occur on the main thread. This makes sure that text cannot be appended out of order. This should fix an inconsistent issue that Khushi ran into.

Also clean up sleeping, to make things look a bit nicer.


ErisMik (approved):

Gabe-Picovoice:
@ErisMik Thanks for reading through this PR! Sorry, I made one last set of changes.

I was organizing this demo to be closer to how Matt did call-assist, then I noticed a race-condition w/ the animation thread that was probably the result of the last issue khushi and I saw. Ran the demo locally and it looks good.

ErisMik (approved):

PR #144, opened 2026-05-20
Speaker Aware Voice Assistant [Android]

ErisMik:


matt200-ok (commented):
When you remove a user profile then try to create an admin profile it creates a user profile instead of an admin profile.

recipes/speaker-aware-voice-assistant/android/README.md
```diff
@@ -0,0 +1,68 @@
...
+[Rhino Speech-to-Intent](https://picovoice.ai/docs/rhino/) to the assets folder.
+
+Lastly, it will place your `AccessKey` from Picovoice Console into the `ACCESS_KEY` variable
+in [MainActivity.java](speaker-aware-voice-assistant/src/main/java/ai/picovoice/speakerawarevoiceassistant/MainActivity.java).
+
+```console
+python setup.py \
+    --access_key ${ACCESS_KEY} \
+    --keyword_path ${PATH_TO_PPN} \
+    --context_path ${PATH_TO_RHN} \
```
Gabe-Picovoice:
```suggestion
    --context_path ${PATH_TO_RHN}
```

recipes/speaker-aware-voice-assistant/android/speaker-aware-voice-assistant/src/main/java/ai/picovoice/speakerawarevoiceassistant/MainActivity.java
```diff
@@ -0,0 +1,752 @@
...
+
+            orca = new Orca.Builder()
+                    .setAccessKey(ACCESS_KEY)
+                    .setModelPath(ORCA_MODEL_FILE)
+                    .build(this);
+
+            eagle = new Eagle.Builder()
+                    .setAccessKey(ACCESS_KEY)
+                    .setVoiceThreshold(0.1f)
+                    .build(this);
```
Gabe-Picovoice: When I press "start testing", I notice 2 - 3s of latency. It would be great if we could only load these once at startup. If not, having a loading spinner here would be nice.

Also, I believe the `setOnClickListener` callback comes from the main UI thread, so we'd want loading to be separate to not block UI updates.

recipes/speaker-aware-voice-assistant/android/speaker-aware-voice-assistant/src/main/java/ai/picovoice/speakerawarevoiceassistant/MainActivity.java
```diff
@@ -0,0 +1,752 @@
...
+    USER
+}
+
+public class MainActivity extends AppCompatActivity {
+    private static final String TAG = "PICOVOICE";
+    private static final String ACCESS_KEY = "${YOUR_ACCESS_KEY_HERE}";
+    private static final String WAKE_WORD_FILE = "keyword.ppn";
+    private static final String CONTEXT_FILE = "context.rhn";
+    private static final String ORCA_MODEL_FILE = "orca_params_en_female.pv";
+    private static final float EAGLE_THRESHOLD = 0.75f;
```
Gabe-Picovoice: My phone microphone gave me scores of 0.45 - 0.57. I was able to get above 0.75 but it took ~10 tries. I suspect the reason is because the wakeword is so short compared to the length of the longer commands (plus the phone microphone being worse).

Gabe-Picovoice (changes requested):
Looks good for the most part, and works well with my voice!

Design notes:
- The main title text should be centered when multi-line (on my device it wraps to two lines, then aligns left). This also happens to the enrolled profile pellets.
- A newline between "Try one of the following commands:" and the options would be nice. Same with "Try the following phrases:". Alternatively, you could remove them both. Changing the prose be "here is what you should do" rather than "consider this" could make it less confusing.
- When rhino fails to understand a command, it should not speak, refresh the buffer, and continue listening. This is how the python version was implemented, and allows for users to try a few times to speak their command.

matt200-ok (approved):

recipes/speaker-aware-voice-assistant/android/speaker-aware-voice-assistant/src/main/java/ai/picovoice/speakerawarevoiceassistant/MainActivity.java
ErisMik: I lowered the threshold to 0.5 :)

recipes/speaker-aware-voice-assistant/android/speaker-aware-voice-assistant/src/main/java/ai/picovoice/speakerawarevoiceassistant/MainActivity.java
ErisMik: Added loading spinner and moved engine loading to a separate executor.

ErisMik:
> Looks good for the most part, and works well with my voice!
> 
> Design notes:
> 
>     * The main title text should be centered when multi-line (on my device it wraps to two lines, then aligns left). This also happens to the enrolled profile pellets.
> 
>     * A newline between "Try one of the following commands:" and the options would be nice. Same with "Try the following phrases:". Alternatively, you could remove them both. Changing the prose be "here is what you should do" rather than "consider this" could make it less confusing.
> 
>     * When rhino fails to understand a command, it should not speak, refresh the buffer, and continue listening. This is how the python version was implemented, and allows for users to try a few times to speak their command.

Comments addressed except for the last one about the rhino failure behavior. Unless I am mistaken the behavior you describe here is NOT what the python demo does? Perhaps I am just not understanding.

Gabe-Picovoice:
> Comments addressed except for the last one about the rhino failure behavior. Unless I am mistaken the behavior you describe here is NOT what the python demo does? Perhaps I am just not understanding.

Ah, my bad! I added this in my demo, but the python demo doesn't have it so it's all good.

Gabe-Picovoice (approved):
Looks good! The spinners turned out well.

PR #145, opened 2026-05-20
Document QA [WEB]

ErisMik:


recipes/document-qa/web/public/index.html
```diff
@@ -0,0 +1,76 @@
...
+      <div id="main-container" style="opacity: 0;">
+        <div id="chat-block-container">
+          <div id="chat-block">
+          </div>
+
+          <div id="volume-meter-user" style="opacity: 0;">
+            <div class="volume-bar" id="bar1"></div>
+            <div class="volume-bar" id="bar2"></div>
+            <div class="volume-bar" id="bar3"></div>
+          </div>
```
matt200-ok: When stopping the demo during speech the VU meter is still visible even when the demo is not listening.

recipes/document-qa/web/public/controller.js
```diff
@@ -0,0 +1,328 @@
...
+    loadContainerParent.style.visibility = "hidden";
+  });
+
+  initButton.addEventListener("click", async () => {
+    if (documentFile.disabled) {
+      stopBubbleDot();
+      await resetDemo();
+    } else {
+      initButton.innerText = "Stop Demo"
+      documentFile.disabled = true;
```
matt200-ok: I think it would be good to introduce a dedicated skip speech button that does not stop the demo.

recipes/document-qa/web/src/index.ts
```diff
@@ -0,0 +1,390 @@
...
+      if (pcm !== null) {
+        object!.audio.stream(pcm);
+
+        pcmBuffered += pcm.length
+        if (pcmBuffered >= 16000) {
+          object!.audio.play();
+        }
+      }
+      streamRelease();
+    }
```
Gabe-Picovoice, line 286: I notice a second or two of delay between when the text starts displaying and when orca starts speaking. I think it would be best if the two displayed at the same time, even though it will increase the apparent text output latency.

I'm not sure whether simply moving sendMessage after `synthesize` will solve this problem, or if we need to buffer both pcm and text.

recipes/document-qa/web/src/audio_stream.ts
```diff
@@ -0,0 +1,83 @@
...
+    streamSource.start();
+    this._isPlaying = true;
+
+    streamSource.onended = (): void => {
+      this._isPlaying = false;
+      if (this._audioBuffers.length > 0) {
+        this.play();
+      }
+    };
+  }
```
Gabe-Picovoice, line 48: The demo works well, although I do notice some consistent crackling around the beginning of large responses that sometimes occurs in the middle of words.

I think what is happening is that in-between the stream ending and `onended` being triggered causes the the stream to stop playing audio. Since orca is in streaming mode, I suspect it is possible for orca to only output part of a word, which is why we notice crackling. If these pauses occurred between words, I would expect to hardly notice. (Although it would still be less than ideal.)

I think https://developer.mozilla.org/en-US/docs/Web/API/AudioScheduledSourceNode is not designed for streaming audio, so it might be a good idea to use [web streaming API](https://googlechromelabs.github.io/web-audio-samples/audio-worklet/basic/hello-audio-worklet/). Although if that seems like too much of a pain, you could probably time it so that these gaps only ever occur between words by disabling orca streaming, and a user would probably not notice.

Gabe-Picovoice (changes requested):
It looks like this demo used the old call-screen demo as a reference. It would be great if we could update the UI in the same way [this PR updates call-screen](https://github.com/Picovoice/pico-cookbook/commit/ba89f2bf3f5c28b32fdb66dfab37d86471470c31#diff-90724da47e3da58a52a101d824dd5e0eb0120c61dc77c58c013f7aae973d8df2), since it includes a lot of Ian's suggestions that apply to this demo as well (centered button, only one status label, ending the demo to change the document brings you back to the start page, etc..).

recipes/document-qa/web/src/index.ts
ErisMik: As discussed, moved call to reduce latency slightly but kept as is.

ErisMik:
> It looks like this demo used the old call-screen demo as a reference. It would be great if we could update the UI in the same way [this PR updates call-screen](https://github.com/Picovoice/pico-cookbook/commit/ba89f2bf3f5c28b32fdb66dfab37d86471470c31#diff-90724da47e3da58a52a101d824dd5e0eb0120c61dc77c58c013f7aae973d8df2), since it includes a lot of Ian's suggestions that apply to this demo as well (centered button, only one status label, ending the demo to change the document brings you back to the start page, etc..).

I applied a bunch of the changes as seen here. Demo looks a lot cleaner now.

recipes/document-qa/web/public/index.html
```diff
@@ -0,0 +1,62 @@
...
+
+    <div id="main-container-parent" style="visibility: hidden;">
+      <div id="main-container" style="opacity: 0;">
+        <h1 style="text-align: center; margin: 15px 0 40px 0;">Document QA</h1>
+
+        <div style="display: flex; flex-direction: row; justify-content: center; height: 100%;">
+
+          <div id="chat-block-container">
+
+            <div id="chat-block">
```
matt200-ok, line 35: It seems like the dots don't always reappear and the demo wasn't scrolling to the bottom for me.

matt200-ok (approved):

recipes/document-qa/web/public/style.css
```diff
@@ -0,0 +1,431 @@
...
+    text-align: center;
+    font-weight: bolder;
+    font-size: 16px;
+
+    color: #222;
+}
+#chat-block {
+    width: 100%;
+
+    overflow-y: scroll;
```
Gabe-Picovoice:
```suggestion
    overflow-y: auto;
```

recipes/document-qa/web/public/controller.js
```diff
@@ -0,0 +1,294 @@
...
+      currentBubbleText = text;
+      chatBlock.appendChild(bubble);
+      chatBlock.scrollTop = chatBlock.scrollHeight;
+
+    } else if (message === "NEW_AI_BUBBLE") {
+      let text = obj;
+      let bubble = document.createElement("div");
+      bubble.classList.add("ai-bubble");
+      bubble.style.opacity = "0";
+      bubble.innerHTML += text;
```
Gabe-Picovoice: The model was outputting information about header files and it gets hidden because the text is treated as html. My bad for using innerHTML originally!

```suggestion
      bubble.innerText += text;
```

recipes/document-qa/web/src/index.ts
```diff
@@ -0,0 +1,415 @@
...
+      sendMessage("ADD_TO_BUBBLE", transcript.transcript);
+    }
+
+    if (transcript.isEndpoint) {
+      sendMessage("ADD_TO_BUBBLE", " ");
+      const cheetah = object!.cheetah;
+      cheetah.flush();
+    }
+
+    let userText = makeRequest("BUBBLE_CONTENTS").replace("[USER] ", "").trim();
```
Gabe-Picovoice: Looks like the dots are part of the text too. I find that in noisy environments, cheetah tends to hit endpoints frequently.

```suggestion
    let userText = makeRequest("BUBBLE_CONTENTS").replace("[USER] ", "").replace(".", "").trim();
```

Gabe-Picovoice (commented):
Looks great! Just a few small improvements.

Gabe-Picovoice (approved):

PR #148, opened 2026-05-22
Voice picking [Web]

Gabe-Picovoice:


recipes/voice-picking/web/public/controller.js
```diff
@@ -0,0 +1,235 @@
...
+
+    btnInit.disabled = true;
+
+    try {
+      if (firstStart) {
+        await Picovoice.init(accessKey, callbacks);
+        firstStart = false;
+      }
+
+      accessKeyInput.classList.add('hidden');
```
matt200-ok, line 243: The demo does not scroll back to the top when you restart it.

recipes/voice-picking/web/src/states.ts
```diff
@@ -0,0 +1,506 @@
...
+export abstract class State {
+    toString(): string {
+        return self.constructor.name;
+    }
+
+    static create(state: RecipeStates, step: Step): State {
+        switch (state) {
+            case RecipeStates.STANDBY:
+                return new RecipeStandbyState(step as PorcupineStep);
+            case RecipeStates.TASK_LOCATION_PROMPT:
```
matt200-ok, line 82: When the check digits are wrong the prompt should change saying that the check digits did not match. Try to look through the rest of the demo to see if any of the other prompts are supposed to change.

Gabe-Picovoice: I noticed that python demo displays two lines, while the the voice guided field demo only displays one of them. I chose to only display the second. I'll change the wording so that it uses both lines but in a shorter excerpt.

Gabe-Picovoice: I ended up just speaking both in sequence, since the text is already pretty terse, and I think it's better this way.

recipes/voice-picking/web/src/steps.ts
```diff
@@ -0,0 +1,312 @@
...
+        onSynthesis?: (alignments: OrcaAlignment[]) => void,
+    ) {
+        try {
+            const { pcm, alignments } = await this.orca.synthesize(prompt);
+            if (onSynthesis) {
+                onSynthesis(alignments);
+            }
+
+            this.audio.stream(pcm);
+            this.audio.play();
```
matt200-ok, line 104: The cancel button currently does nothing when audio is playing but it should be able to cancel the audio.

recipes/voice-picking/web/public/controller.js
```diff
@@ -0,0 +1,244 @@
...
+    },
+
+    createCard,
+    setActiveCard,
+    setCompletedCard,
+    setCardValue,
+    goToInitScreen,
+  };
+
+  btnInit.onclick = async () => {
```
matt200-ok, line 222: Every time the demo starts the cards should be in the same position scrolled all the way to the top. Make sure that if you start the demo, scroll to the bottom, stop the demo, then restart the demo it should be scrolled to the top.

recipes/voice-picking/web/src/steps.ts
Gabe-Picovoice: Interrupting the audio stream didn't end up being as hard as I thought, so good suggestion!

recipes/voice-picking/web/public/controller.js
Gabe-Picovoice: Ty! I really should have tested my fix for this; it's working properly now.

matt200-ok (approved):

recipes/voice-picking/web/src/steps.ts
```diff
@@ -0,0 +1,315 @@
...
+                console.log("on synthesis");
+                onSynthesis(alignments);
+            }
+
+            this.audio.stream(pcm);
+            this.audio.play();
+
+        } finally {
+            await this.audio.waitPlayback(() => { console.log(isRunning); return !isRunning; });
+            console.log("AFTER!!!!!!!!1");
```
laves: remove and double-check there aren't any other debug logs

recipes/voice-picking/web/src/index.ts
```diff
@@ -0,0 +1,155 @@
...
+
+const init = async (accessKey: string, cb: DemoCallbacks): Promise<void> => {
+  updateCallbacks(cb);
+
+  callbacks.clearStatus();
+
+  let i = 0;
+  for (const task of TASKS) {
+    callbacks.createCard(`location-${i}`, "CONFIRM LOCATION", `Location: ${task.checkDigit}`);
+    callbacks.createCard(`pick-${i}`, "PICK ITEM", `Item: ${task.itemName} (${task.quantity})`);
```
laves: I'm not sure about this extra text in the upper right corner. I wasn't sure if they were examples of how I should respond or what and it lent to several rejects as I tried to say 'blue widgets 3'. I don't think we really need it

recipes/voice-picking/web/src/states.ts
```diff
@@ -0,0 +1,521 @@
...
+
+        if (inference && inference.isUnderstood && RecipeTaskPickReportState.VALID_INTENTS.includes(inference.intent!)) {
+            if (inference.intent == 'exitWorkflow') {
+                return {
+                    outcome: inference,
+                    next: {
+                        state: RecipeStates.COMPLETE_PROMPT,
+                        prompt: "Picking workflow ended.",
+                    }
+                };
```
laves, line 424: I know the python demo doesn't make this clear, but if you could have some text somewhere that indicates that there is a command to end the workflow, that would be nice. One run-through I was halfway through and it miss identified what I said as `exitWorkflow` and I was pretty thrown by the demo suddenly ending

recipes/voice-picking/web/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+These instructions assume your current working directory is `recipes/voice-picking/web`.
+
+### 1. Train a Wake Word Model
+
+1. Open [Picovoice Console](https://console.picovoice.ai/)
+2. Go to Porcupine Wake Word.
+3. Enter your desired wake phrase.
+4. Click Train.
+5. Select your target platform and download the generated wake word model file (`.ppn`).
```
laves:
```suggestion
5. Select target platform **Web (WASM)** and download the generated wake word model file (`.ppn`).
```

recipes/voice-picking/web/README.md
```diff
@@ -0,0 +1,91 @@
...
+Save the downloaded file somewhere accessible on your machine. You will pass its path to the demo with `--porcupine_keyword_path`.
+
+### 2. Train the Speech-to-Intent Model
+
+1. Open [Picovoice Console](https://console.picovoice.ai/)
+2. Go to Rhino Speech-to-Intent.
+3. Create an empty Rhino context.
+4. Click Import YAML in the top-right corner.
+5. Paste the [Rhino context YAML](../res/context.yml) for this demo.
+6. Download the generated Rhino context file (`.rhn`) for your target platform.
```
laves:
```suggestion
6. Download the generated Rhino context file (`.rhn`) for target platform **Web (WASM)**.
```

laves (changes requested):
Overall, nicely done! Be sure to test on Safari if you haven't already. I've just been through some pain on our website debugging Safari-specific bugs, so I want to encourage us to test in Safari more before we ship demos

recipes/voice-picking/web/src/states.ts
Gabe-Picovoice: Ah, yeah I ran into that from rhino as well, since "item damaged" is really close to "i am done". Under the instructions header, I included 

```
- When asked to pick a number of items, respond with:
  - `picked ${N}`, where `N` is the number of items you picked
...
  - `I am done` to exit the workflow early
- See `../res/context.yml` for more details
```

Is this sufficient?

laves: I was thinking something in the UI would be even better so that if they haven't read the docs super carefully they still can understand why the demo ended early

Gabe-Picovoice: Ah, I see. I suppose the other demos don't have this feature so it makes sense for this one to be different. I thought of a good way to convey this.

laves (approved):

PR #156, opened 2026-05-27
Voice Picking [Android]

Gabe-Picovoice:


recipes/voice-picking/android/voice-picking/src/main/java/ai/picovoice/voicepicking/MainActivity.java
```diff
@@ -0,0 +1,1148 @@
...
+import ai.picovoice.android.voiceprocessor.VoiceProcessorException;
+import ai.picovoice.koala.Koala;
+import ai.picovoice.orca.Orca;
+import ai.picovoice.orca.OrcaAudio;
+import ai.picovoice.orca.OrcaSynthesizeParams;
+import ai.picovoice.porcupine.Porcupine;
+import ai.picovoice.rhino.Rhino;
+import ai.picovoice.rhino.RhinoInference;
+
+public class MainActivity extends AppCompatActivity {
```
matt200-ok, line 54: The demo skips the proper response when you report an issue in the last step. It always says picking workflow complete even if you report a short pick. If you report a short pick anywhere else it gives a special message and the final bin should be the same.

Gabe-Picovoice: This is how the python version is implemented as well. I suppose the original reason was brevity?
```
if next_task_index >= len(tasks):
    if intent == 'confirmPickedQuantity':
        text = f"Recorded picked {slots['quantity']}."
    elif intent == 'reportShortPick':
        text = f"Recorded short pick {slots['quantity']}."
    elif intent == 'reportDamagedItem':
        text = "Recorded damaged item."
    else:
        text = "Recorded empty location."
  
    sleep(.1)
    event.set()
    thread.join()
  
    return Transition(
        outcome=inference,
        next_state=RecipeStates.COMPLETE_PROMPT,
        next_state_kwargs={
            'tasks': tasks,
            'task_index': next_task_index,
        })
```

matt200-ok (approved):

laves (approved):

PR #160, opened 2026-05-28
Speaker Aware Wake Word demo for iOS

laves:


matt200-ok (approved):

recipes/speaker-aware-wake-word/ios/README.md
```diff
@@ -0,0 +1,42 @@
...
+### 1. Train a Wake Word Model
+
+Follow instruction on Porcupine Wake Word [documentation](https://picovoice.ai/docs/porcupine/#custom-wake-words) to
+train a wake word model on [Picovoice Console](https://console.picovoice.ai/) in minutes. Ensure you train the wake word for the **iOS** platform.
+
+### 2. Add Wake Word to the Project
+
+Run the setup script to copy the wake word model provided for [Porcupine Wake Word](https://picovoice.ai/docs/porcupine/) to the project folder.
+
+It will also place your `AccessKey` from Picovoice Console and name into the `ACCESS_KEY` variable
```
Gabe-Picovoice:
```suggestion
It will also place your `AccessKey` from Picovoice Console into the `ACCESS_KEY` variable
```

recipes/speaker-aware-wake-word/ios/SpeakerAwareWakeWord/ViewModel.swift
```diff
@@ -0,0 +1,365 @@
...
+                accessKey: ACCESS_KEY,
+                keywordPaths: [keywordPath],
+                sensitivities: [0.5]
+            )
+            eagleProfiler = try EagleProfiler(
+                accessKey: ACCESS_KEY,
+                minEnrollmentChunks: 4,
+                voiceThreshold: 0.0)
+
+            enrollMaxSamples = EagleProfiler.frameLength * 96
```
Gabe-Picovoice: Where is this `96` coming from? I notice that web has a similar, yet different, line https://github.com/Picovoice/pico-cookbook/blob/bd234b426ee47f79d2f4ac87fe8bdec2af6da437/recipes/speaker-aware-wake-word/web/src/index.ts#L138

Gabe-Picovoice (commented):
- I notice when enrolling, the progress bar does not animate for the last arc of the circle. Instead, the screen immediately changes as enrolling completes.

recipes/speaker-aware-wake-word/ios/SpeakerAwareWakeWord/ViewModel.swift
laves: This is setting the max amount of data to buffer while waiting for the keyword to be spoken. In mine, I'm setting it to align with Eagle's frame size and the 96 is just a threshold where we starting throwing away frames. I could make the math look more complex to express it in seconds, but I thought is was sufficient to just set a value for it in frames. I could hoist the magic number as a constant though

Gabe-Picovoice: ah, I see. I think it's fine as it is. Good to know both versions were intentional

Gabe-Picovoice (approved):

PR #165, opened 2026-05-30
Voice Picking [iOS]

Gabe-Picovoice:


recipes/voice-picking/ios/setup.py
```diff
@@ -103,18 +99,18 @@ def main() -> None:
...
+        '--rhino_context_path',
         required=True,
         help='Absolute path to the Rhino model file (`.rhn`).')
     args = parser.parse_args()
 
     public_folder = os.path.join(
         os.path.dirname(__file__),
-        "VoiceGuidedFieldReporting",
+        "VoicePicking",
         "resources")
```
matt200-ok, line 114: Missing gitkeep

recipes/voice-picking/ios/VoicePicking/MainView.swift
```diff
@@ -74,16 +96,19 @@ struct MainView: View {
...
+                            hasAlternate: cardData.hasAlternate,
+                            isAlternateActive: cardData.isAlternateActive)
+                    }.onChange(of: viewModel.cardData, {
                         if viewModel.activeCard != nil {
                             proxy.scrollTo(
                                 viewModel.activeCard!,
-                                anchor: .center)
+                                anchor: .bottom)
                         }
                     })
```
matt200-ok: The demo was not scrolling for me.

recipes/voice-picking/ios/VoicePicking/MainView.swift
```diff
@@ -63,6 +65,26 @@ struct CardView: View {
...
+
+            if hasAlternate {
+                VStack {
+                    HStack {
+                        Text("or EXIT WORKFLOW")
+                            .frame(maxWidth: .infinity, alignment: .center)
+                            .multilineTextAlignment(.center)
+                            .foregroundStyle(isAlternateActive ? .blue : .darkGray)
+                            .bold()
+                        Spacer()
```
matt200-ok: I don't think this should use the same UI style as the report fields because it cannot be filled in. Also instead of saying exit workflow I think it should say stop picking so people know what to say to stop.

recipes/voice-picking/ios/.gitignore
```diff
@@ -24,5 +24,6 @@ Pods
 # Exclude picovoice cloned repositories and resources
 cheetah
 orca
-koala
-VoiceGuidedFieldReporting/resources
\ No newline at end of file
+rhino
+porcupine
+VoicePicking/resources
```
laves: the only two repos cloned are orca and koala. Cheetah, rhino and porcupine can be removed.

recipes/voice-picking/ios/VoicePicking/ViewModel.swift
```diff
@@ -0,0 +1,906 @@
...
+        orcaStep: OrcaStep,
+        porcupineStep: PorcupineStep,
+        rhinoStep: RhinoStep
+    ) async {
+        await viewModel.initCards()
+
+        states = [
+            .STANDBY: StandbyState(
+                viewModel: viewModel,
+                porcupineStep: porcupineStep,
```
laves: all of these state objects have trailing `,` without a following argument, which causes a build error on my Xcode version. They relaxed this syntax requirement in more recent versions, but I think we should fix it so that it has wider version support (my version 16.2 is not even two years old)

laves (commented):
be sure to test on a real device if you haven't

laves (approved):

recipes/voice-picking/ios/VoicePicking/MainView.swift
Gabe-Picovoice: Good catch, ty! I'll test the iOS demos horizontally from now on \:)

PR #170, opened 2026-06-03
Self checkout [Python]

Gabe-Picovoice:


recipes/self-checkout/python/main.py
```diff
@@ -0,0 +1,1023 @@
+import sys
```
matt200-ok, line 1: The demo is not recognizing a number of the commands such as `Pay` for me. Make sure you validate that every variant of each command is being handled correctly.

Gabe-Picovoice: I noticed `Pay` was difficult, although I was able to trigger it with a bit of effort. I've increased sensitivity which appears to help a little. I also removed the hint to say some of the shorter commands that were difficult.

matt200-ok: I find the demo misfires often. You might need to reduce the amount of variation in the commands so that they don't overlap. The most common misfire was going to pay when I tried to scan items or ask for the total.

Gabe-Picovoice: I think "total" and "that's all" were a bit too similar for one of the cases you mentioned. In any case, since we display a list of options, I'll make sure these instances have fewer additional wordings.

I tested a bunch of times and it seems better now.

ksyeo1010 (approved):

matt200-ok (approved):

PR #173, opened 2026-06-04
 Update Call Assist prompt

Gabe-Picovoice:


Gabe-Picovoice:
@khushi-nakra changes look good!

PR #177, opened 2026-06-05
Food ordering [Android]

Gabe-Picovoice:
.


recipes/food-ordering/android/food-ordering/src/main/java/ai/picovoice/foodordering/MainActivity.java
```diff
@@ -0,0 +1,1393 @@
+/*
```
matt200-ok, line 1: I think having a tooltip or suggestion would be good so people know what they could say. I think you can just show the tooltip when their order is empty.

Gabe-Picovoice: Good suggestion! Will do

recipes/food-ordering/android/food-ordering/src/main/java/ai/picovoice/foodordering/MainActivity.java
```diff
@@ -0,0 +1,1413 @@
...
+                            CardUI activeCard = cards.get(index);
+                            scrollView.smoothScrollTo(0, activeCard.root.getTop());
+
+                            runOnUiThread(() -> {
+                                cards.remove(index);
+                                reportContainer.removeViewAt(index);
+
+                                if (cards.size() == 0) {
+                                    toggleEmptyOrderSuggestion();
+                                }
```
ksyeo1010: do we need nested run on ui thread blocks?

matt200-ok (approved):

recipes/food-ordering/android/food-ordering/src/main/java/ai/picovoice/foodordering/MainActivity.java
```diff
@@ -0,0 +1,1413 @@
...
+                    .start();
+            isRunning = false;
+        });
+    }
+
+    class CardUI {
+        View root;
+        View leftContainer;
+        TextView valueView;
+    }
```
ksyeo1010: we dont need every class in a single file

recipes/food-ordering/android/food-ordering/src/main/java/ai/picovoice/foodordering/MainActivity.java
```diff
@@ -0,0 +1,1413 @@
...
+
+    private static final String ACCESS_KEY = "${YOUR_ACCESS_KEY_HERE}";
+    private static final String KEYWORD_MODEL = "food_ordering_android.ppn";
+    private static final String CONTEXT_MODEL = "food_ordering_android.rhn";
+
+    private static final String TTS_MODEL = "orca_params_en_female.pv";
+    private static final String NS_MODEL = "koala_params.pv";
+
+    private static final int INACTIVE_COLOUR = Color.parseColor("#7f8c8d");
+    private static final int ACTIVE_COLOUR = Color.parseColor("#377dff");
```
ksyeo1010: we have resources file for colors

recipes/food-ordering/android/food-ordering/src/main/java/ai/picovoice/foodordering/MainActivity.java
Gabe-Picovoice: Ah, good catch! I got it backwards here. Scrolling to a newly created item is the only operation that needs to occur one frame late, since it will fail to scroll if the object didn't exist at the start of the UI frame.

recipes/food-ordering/android/food-ordering/src/main/java/ai/picovoice/foodordering/MainActivity.java
Gabe-Picovoice: Good suggestion, I've refactored the steps, card UI, and order models into their own files. I left all the state machine states since it's contains the main logic of the demo.

recipes/food-ordering/android/food-ordering/src/main/java/ai/picovoice/foodordering/MainActivity.java
Gabe-Picovoice: Good catch, looks like these were just dead code. I also removed a few unnecessary imports I could find.

ksyeo1010 (approved):

